PR #197, opened 2026-04-09
fix(mindmap): route HTML wrapper generation through MindMapHtmlWrapper + Playwright tests (#196)

### LINKED ISSUE #196: feat(mindmap): fix + publish + Playwright-test HTML wrapper generation (regression of #129)
## Context

Issue #129 closed as done, but the pipeline actually produces **broken HTML wrappers** and none are published alongside the committed SVGs in [Cards/Fallacies/Mindmaps/](Cards/Fallacies/Mindmaps/). Current state observed on ai-01 after a pipeline run with `Mode = Mindmapper`.

### 1. `[SVGCONTENT]` placeholder never replaced (included variant broken)

The template [included.html:393](Cards/Fallacies/Mindmaps/included.html#L393) uses the placeholder `[SVGCONTENT]` to embed the inlined SVG. But the generator subclasses [FallacyMindMapDocumentConfig.cs:1324](Generation/Converters/Argumentum.AssetConverter/Mindmapper/FallacyMindMapDocumentConfig.cs#L1324) and [VirtueMindMapDocumentConfig.cs:894](Generation/Converters/Argumentum.AssetConverter/Mindmapper/VirtueMindMapDocumentConfig.cs#L894) look for the string ``<!-- Insert here the SVG -->`` which is **not** in the template. Result: every generated `Fallacies_{lang}.html` and `Argumentation_Virtues_{lang}.html` contains the literal text `[SVGCONTENT]` instead of the inlined SVG. The included variant is visually broken.

Evidence (generated file from last run):
~~~
Fallacies_fr.html:393:        [SVGCONTENT]
~~~

The base class [MindMapDocumentConfig.cs:785](Generation/Converters/Argumentum.AssetConverter/Mindmapper/MindMapDocumentConfig.cs#L785) uses the correct `[SVGCONTENT]` placeholder — the regression is in the two subclass overloads.

### 2. Virtues wrapper name hardcoded to `fr` for all languages

[VirtueMindMapCreatorConfig.cs:58](Generation/Converters/Argumentum.AssetConverter/Mindmapper/VirtueMindMapCreatorConfig.cs#L58) sets ``DocumentName = ""Argumentation_Virtues_fr.html""`` — no `[LANGUAGE]` placeholder. As a result every language outputs a file literally named `Argumentation_Virtues_fr.html` (the `Replace(""[LANGUAGE]"", language)` at [VirtueMindMapDocumentConfig.cs:880](Generation/Converters/Argumentum.AssetConverter/Mindmapper/VirtueMindMapDocumentConfig.cs#L880) is a no-op on this string). File-level collision is avoided only because each language has its own output directory.

Fallacies uses `Fallacies_[LANGUAGE].html` correctly at [FallacyMindMapCreatorConfig.cs:80](Generation/Converters/Argumentum.AssetConverter/Mindmapper/FallacyMindMapCreatorConfig.cs#L80) — so the fix is symmetric.

### 3. Wrappers are never published to the source tree

The 20 Fallacy/Virtue SVGs (Batik) were committed by PR #184 to `Cards/Fallacies/Mindmaps/{fr,en,ru,pt}/`. The wrappers the pipeline generates land in `bin/Debug/.../Target/{lang}/Documents/` (gitignored). There is no copy-back step, so **zero HTML wrappers are actually checked in**. The website and validation workflows have nothing to link to.

### 4. Orphan wrapper with hash suffix

A stale `Fallacies_fr_5893947243594948969.html` lingers in the Target directory from an older buggy run (filename suggests a `GetHashCode()` leakage). Not directly a bug in the current code path, but an indicator the naming logic has drifted over time.

## Gap against the user requirement (2026-04-09)

The wrappers are supposed to be directly testable in Playwright so the team can visually verify what ships (cf. the \"chemins des nouveaux doc parfaitement MAJ\" feedback on 2026-04-09). Today there is no test at all on the wrapper generation — zero Playwright coverage, and the broken placeholder is not caught by any unit test.

## Scope

Phase 1 — code fixes:
- [ ] Fix `[SVGCONTENT]` placeholder in [FallacyMindMapDocumentConfig.cs:1324](Generation/Converters/Argumentum.AssetConverter/Mindmapper/FallacyMindMapDocumentConfig.cs#L1324) and [VirtueMindMapDocumentConfig.cs:894](Generation/Converters/Argumentum.AssetConverter/Mindmapper/VirtueMindMapDocumentConfig.cs#L894) to match the template string.
- [ ] Replace hardcoded `Argumentation_Virtues_fr.html` in [VirtueMindMapCreatorConfig.cs:58](Generation/Converters/Argumentum.AssetConverter/Mindmapper/VirtueMindMapCreatorConfig.cs#L58) and line 65 with `Argumentation_Virtues_[LANGUAGE].html` / `_ext.html`, mirroring Fallacies.
- [ ] Add a publish/copy step that writes the four Fallacies wrappers + four Virtues wrappers to `Cards/Fallacies/Mindmaps/{lang}/` when `OverwriteExistingHtmlMaps = true` (so editors can commit them into source). Non-destructive by default (skip existing).
- [ ] Remove the stale `Fallacies_fr_5893947243594948969.html` orphan cleanup (investigate one-time root cause before rm).

Phase 2 — Playwright tests (the main deliverable):
- [ ] Add `Argumentum.AssetConverter.VisualTests/MindmapWrapperTests.cs` that, for each `(lang, variant) in {(fr|en|ru|pt) × (included|external)}`:
  1. Loads the generated HTML wrapper file via `file://` in Playwright Chromium.
  2. Asserts the page has a `#mindmap` container that is non-empty (for included: inline `<svg>`; for external: loaded `<object id=\"svgObject\">` with `data` attribute pointing at an existing file).
  3. Asserts there is at least one `.node` element in the SVG document (or content document for external).
  4. Clicks a node and asserts the overlay `<card>` loses the `hidden` class + `.famille`/`.sous_famille` are populated.
  5. Takes a screenshot for Verify (snapshot).
- [ ] Test runs from `dotnet test` without a full pipeline pre-run (fixture wrappers committed to test data, or wrappers regenerated as part of the test fixture).
- [ ] CI-friendly: no dependency on Freeplane/FreeMind GUI (tests operate on committed SVG + HTML artefacts).

Phase 3 — Ops:
- [ ] Regenerate the 16 wrappers (4 langs × 4 files: Fallacies included + ext, Virtues included + ext) and commit them to `Cards/Fallacies/Mindmaps/{lang}/`. This run is light (no harvest, no PDF).
- [ ] Update [CLAUDE.md](CLAUDE.md) pipeline status section to note the HTML wrappers are published + Playwright-tested.

## Definition of Done

1. `dotnet test` runs MindmapWrapperTests green (all 8 wrappers pass the Playwright checks).
2. `git status` shows 16 committed HTML wrappers in `Cards/Fallacies/Mindmaps/{lang}/` (8 Fallacies + 8 Virtues).
3. Opening `Cards/Fallacies/Mindmaps/fr/Fallacies_fr.html` in a fresh Chromium shows the inline SVG with clickable nodes and the overlay card — manually verified once by ai-01 before the PR merges.
4. No `[SVGCONTENT]` literal remaining in any committed wrapper file.

## Related

- Closes the gap left open by #129 (which declared the feature complete prematurely).
- Provides the visual QA substrate needed to make progress on #140 (multilingual visual verification).
- Unblocks the website publication path (#131/#132).

## Assignment

- ai-01: all phases (visual QA is ai-01 scope).
- po-2023: out of scope — continues on #193 (GSheet sync).

jsboige:
Closes #196

## Summary
- Extracts `MindMapHtmlWrapper.FormatWrapper` as a pure, tested helper and routes the three `*MindMapDocumentConfig` subclasses through it. Fixes a shipped regression where `Fallacy`/`Virtue` subclasses searched for a `<!-- Insert here the SVG -->` comment that no longer exists in `included.html`, leaving `[SVGCONTENT]` as a literal placeholder in the generated DOM.
- Fixes `VirtueMindMapCreatorConfig` hardcoded `Argumentation_Virtues_fr.html` filename — now uses the standard `[LANGUAGE]` token so each of the four languages ships under its own name.
- Adds two complementary test layers so the regression cannot ship silently again.

### Unit tests — `Argumentum.AssetConverter.Tests/MindmapGeneration/MindMapHtmlWrapperTests.cs`
7 tests exercising `FormatWrapper` against the real committed `included.html` / `external.html`:
- both placeholders are replaced
- the external template contract (no `[SVGCONTENT]`) is documented and enforced
- idempotency (second call is a no-op)
- null handling for both parameters
- `null` template throws

### Playwright integration tests — `Argumentum.AssetConverter.VisualTests/MindmapWrapperTests.cs`
9 tests that stage wrappers in a temp dir, load them via `file://` in headless Chromium, and verify the DOM:
- 4 parametrized tests (fr/en/ru/pt): embed the real committed Batik SVG via `included.html` wrapper, assert an inline `<svg>` is present under `#mindmap` with non-empty path geometry, and that no raw `[SVGCONTENT]` / `[SVGPATH]` ever reaches the DOM.
- 4 parametrized tests (fr/en/ru/pt): external variant references the SVG via `<object data="...">`, with both wrapper and SVG staged side-by-side mirroring the pipeline layout.
- 1 synthetic-SVG test for the **interactive overlay**: inlines an SVG with `class=\"node\"` markers and the `family`/`description`/`example`/`link` attributes the overlay JS expects, clicks the node, and asserts `card .famille` becomes visible and carries the injected family string (via `textContent`, not `innerText`, to bypass the CSS `text-transform: uppercase`).

The synthetic SVG is the regression harness for the interactive layer. The current committed Batik SVGs are visual-only (no `.node` markers — a tracked gap, not a regression), so a test based on them alone could never prove the click path actually works. The synthetic fixture keeps the overlay test decoupled from the SVG generator's evolution.

## Test plan
- [x] `dotnet test ... --filter MindMapHtmlWrapperTests` → 7/7 pass
- [x] `dotnet test ... --filter MindmapWrapperTests` → 9/9 pass
- [ ] Pipeline run with `Mode = Mindmapper` to regenerate the 16 wrappers under `Cards/Fallacies/Mindmaps/{lang}/` (phase 3, deferred to a follow-up PR so this one stays focused on the code + tests).

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/MindmapGeneration/MindMapHtmlWrapperTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/MindmapGeneration/MindMapHtmlWrapperTests.cs
@@ -1,0 +1,108 @@
+using System.IO;
+using Argumentum.AssetConverter.Mindmapper;
+using FluentAssertions;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.MindmapGeneration
+{
+    /// <summary>
+    /// Pure unit tests for <see cref="MindMapHtmlWrapper.FormatWrapper"/> — exercises the two
+    /// placeholder substitutions on both real templates (<c>included.html</c>, <c>external.html</c>)
+    /// without spinning up Playwright. Issue #196 phase 1 — ensures the regression that shipped
+    /// literal "[SVGCONTENT]" strings in the DOM cannot reappear silently.
+    /// </summary>
+    public class MindMapHtmlWrapperTests
+    {
+        private static readonly string RepoRoot =
+            Path.GetFullPath(Path.Combine(AppContext.BaseDirectory, "..", "..", "..", "..", "..", ".."));
+
+        private static readonly string IncludedTemplatePath =
+            Path.Combine(RepoRoot, "Cards", "Fallacies", "Mindmaps", "included.html");
+
+        private static readonly string ExternalTemplatePath =
+            Path.Combine(RepoRoot, "Cards", "Fallacies", "Mindmaps", "external.html");
+
+        [Fact]
+        public void FormatWrapper_Included_ReplacesSvgContentPlaceholder()
+        {
+            var template = File.ReadAllText(IncludedTemplatePath);
+            template.Should().Contain("[SVGCONTENT]",
+                "pre-condition: the committed template must contain the placeholder this fix targets");
+
+            var result = MindMapHtmlWrapper.FormatWrapper(
+                template,
+                "Fallacies_fr.content.svg",
+                "<svg xmlns=\"http://www.w3.org/2000/svg\" id=\"embedded-test\"><g class=\"node\" id=\"n1\"/></svg>");
+
+            result.Should().NotContain("[SVGCONTENT]",
+                "the regression introduced in #129 shipped wrappers with the literal placeholder in the DOM");
+            result.Should().Contain("id=\"embedded-test\"", "the SVG markup must be inlined inside the wrapper");
+            result.Should().Contain("class=\"node\"", "clickable nodes must be preserved in the inline SVG");
+        }
+
+        [Fact]
+        public void FormatWrapper_External_ReplacesSvgPathPlaceholder()
+        {
+            var template = File.ReadAllText(ExternalTemplatePath);
+            template.Should().Contain("[SVGPATH]");
+
+            var result = MindMapHtmlWrapper.FormatWrapper(
+                template,
+                "Fallacies_fr.content.svg",
+                "<svg/>"); // external template ignores SVGCONTENT — safe no-op
+
+            result.Should().NotContain("[SVGPATH]");
+            result.Should().Contain("data=\"Fallacies_fr.content.svg\"",
+                "the <object> tag should reference the SVG file via the replaced [SVGPATH]");
+        }
+
+        [Fact]
+        public void FormatWrapper_External_TemplateHasNoSvgContentPlaceholder()
+        {
+            // Documents the design: only included.html carries [SVGCONTENT]. This prevents future
+            // edits from accidentally embedding an SVG twice in the external variant.
+            var template = File.ReadAllText(ExternalTemplatePath);
+            template.Should().NotContain("[SVGCONTENT]");
+        }
+
+        [Fact]
+        public void FormatWrapper_HandlesNullSvgContentGracefully()
+        {
+            var template = "<html>[SVGPATH] / [SVGCONTENT]</html>";
+
+            var result = MindMapHtmlWrapper.FormatWrapper(template, "a.svg", svgContent: null!);
+
+            result.Should().Be("<html>a.svg / </html>");
+        }
+
+        [Fact]
+        public void FormatWrapper_HandlesNullSvgPathGracefully()
+        {
+            var template = "<html>[SVGPATH] / [SVGCONTENT]</html>";
+
+            var result = MindMapHtmlWrapper.FormatWrapper(template, svgRelativePath: null!, "<svg/>");
+
+            result.Should().Be("<html> / <svg/></html>");
+        }
+
+        [Fact]
+        public void FormatWrapper_NullTemplate_Throws()
+        {
+            Action act = () => MindMapHtmlWrapper.FormatWrapper(null!, "a.svg", "<svg/>");
+            act.Should().Throw<ArgumentNullException>();
+        }
+
+        [Fact]
+        public void FormatWrapper_Idempotent_SecondCallIsNoOp()
+        {
+            var template = "<html>[SVGPATH] [SVGCONTENT]</html>";
+            var once = MindMapHtmlWrapper.FormatWrapper(template, "a.svg", "<svg/>");
+            var twice = MindMapHtmlWrapper.FormatWrapper(once, "b.svg", "<svg id=\"other\"/>");
+
+            // Guardrail: once the placeholders are gone, re-running the helper must not mutate
+            // the content. Documents the expectation and catches regressions where someone adds
+            // a third placeholder without updating the contract.
+            twice.Should().Be(once);
+        }
+    }
+}
--- a/Generation/Converters/Argumentum.AssetConverter.VisualTests/MindmapWrapperTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.VisualTests/MindmapWrapperTests.cs
@@ -1,0 +1,208 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Argumentum.AssetConverter.Mindmapper;
+using Microsoft.Playwright;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Argumentum.AssetConverter.VisualTests
+{
+    /// <summary>
+    /// Playwright integration tests for mind map HTML wrapper rendering.
+    /// Issue #196: proves that the wrappers generated from the committed templates +
+    /// committed Batik SVGs actually render in a real browser, nodes are clickable,
+    /// and the overlay card reacts to clicks. Every assertion corresponds to something
+    /// the user would spot visually when opening the file.
+    ///
+    /// These tests DO NOT require a full pipeline run — they compose the wrapper in a
+    /// temp directory using the same <see cref="MindMapHtmlWrapper.FormatWrapper"/> path
+    /// the pipeline uses, so a regression in the helper will surface here.
+    /// </summary>
+    public class MindmapWrapperTests : IAsyncLifetime
+    {
+        private readonly ITestOutputHelper _output;
+        private IPlaywright _playwright = null!;
+        private IBrowser _browser = null!;
+        private string _tempDir = null!;
+
+        private static readonly string RepoRoot =
+            Path.GetFullPath(Path.Combine(AppContext.BaseDirectory, "..", "..", "..", "..", "..", ".."));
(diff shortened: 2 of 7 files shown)


jsboige:
## Rigorous self-review — critical pass

I re-read the diff as a reviewer who had not seen the PR before. Findings below, with the most important first.

### 🔴 Factual doc error (FIXED in commit `cf36d108`)

`MindMapHtmlWrapper.cs`'s XML doc claimed:
> `[SVGPATH]` — Also present in included.html for fallback, harmless there.

A `grep` against the committed templates proves the opposite: `[SVGPATH]` appears **only** in `external.html`, `[SVGCONTENT]` **only** in `included.html`. The helper's unconditional-replace behavior is still correct (`string.Replace` on a missing needle is a safe no-op), but the doc was misleading. Rewrote it to describe the actual contract and point at the unit test (`FormatWrapper_External_TemplateHasNoSvgContentPlaceholder`) that pins it.

No behavior change — `FormatWrapper` is untouched.

### 🟡 Scope gap — wrappers still missing in source tree (tracked in #198)

PR #197 fixes the **code** that generates wrappers, and the **tests** that verify the code path. It does **not** publish the 16 actual wrappers to `Cards/Fallacies/Mindmaps/{lang}/`. They only exist in `bin/Debug/.../Target/` which is gitignored. That was always phase 3, but worth calling out explicitly so reviewers don't merge this thinking the GitHub-Pages-served wrappers are now fixed. They aren't yet.

Created #198 to track phase 3 with a full manifest of the 16 expected files (4 langs × 2 templates × 2 mindmap types) and two alternative approaches.

### 🟡 Skip-if-exists guard will bite phase 3

Both subclasses have:
```csharp
if (File.Exists(htmlFileName) && !config.OverwriteExistingHtmlMaps)
{
    Logger.Log($"Skip existing Html SVG Wrapper: {htmlFileName}");
}
```
(`FallacyMindMapDocumentConfig.cs:1312`, `VirtueMindMapDocumentConfig.cs:885`, `MindMapDocumentConfig.cs:773`)

If phase 3 starts from a working tree that already has stale wrappers (e.g. someone regenerated once, committed, then the templates evolved), the pipeline will **silently skip them**. Documented in #198's recommended approach.

### 🟡 Code-smell: 3x duplicated `GenerateHtmlSvgWrappers`

There are three **private static** `GenerateHtmlSvgWrappers` methods: one in `MindMapDocumentConfig` (4-arg, no language), two in `Fallacy/VirtueMindMapDocumentConfig` (5-arg, with language). They don't share a base — `FallacyMindMapDocumentConfig : FallacyDocumentConfigBase`, `VirtueMindMapDocumentConfig : VirtueDocumentConfigBase`, `MindMapDocumentConfig : DocumentConfig`. The 4-arg variant is only called from `MindMapCreatorConfig` which isn't wired into the pipeline entrypoint (dead code).

**Not fixing in this PR** — the code smell predates #196 and the fix would fan out into a cross-cutting refactor. Keeping the three sites consistent (all routing through `MindMapHtmlWrapper.FormatWrapper`) is the minimum responsible change.

### 🟡 `RepoRoot` computation is brittle

Both test files compute:
```csharp
Path.GetFullPath(Path.Combine(AppContext.BaseDirectory, "..", "..", "..", "..", "..", ".."))
```
Six `..` is correct for `bin/Debug/net9.0-windows/` → repo root, but any future nesting (publish/, multiple TFMs via MSBuild-item-metadata path tweaks) breaks every test. The more robust pattern is `[CallerFilePath]` to a constant in an anchor file. **Deferred** — affects many test files, out of scope for #197.

### 🟢 CI never runs these tests

Only workflow in `.github/workflows/` is `static.yml` which deploys CardPen to GitHub Pages. **No test runner CI**. This means:
- `MindMapHtmlWrapperTests` runs only when someone locally executes `dotnet test ... --filter MindMapHtmlWrapperTests`
- `MindmapWrapperTests` additionally requires `playwright install` locally

Not a regression from this PR — same situation for all existing tests. Worth a future CI issue but not blocking.

### 🟢 What's strong in this PR

- **Pure helper extraction is clean**: 34 lines including doc, no dependencies, trivial to test.
- **Test layering is the right shape**: 7 unit tests for the substitution logic + 9 Playwright tests for the DOM-level behavior. No overlap, no redundancy.
- **The synthetic SVG for the overlay click test** is the right call — coupling it to the current Batik SVGs (which intentionally have no `.node` markers) would prove nothing about the JS click handler. The comment in `MindmapWrapperTests.cs:112-117` documents this explicitly.
- **The `textContent` vs `innerText` fix** I applied after the first Playwright run (CSS `text-transform: uppercase` was returning `INSUFFISANCE` instead of `Insuffisance`) is a subtle gotcha that will save future contributors an hour of debugging.
- **Null-handling tests** (`FormatWrapper_HandlesNullSvgContentGracefully`, `FormatWrapper_HandlesNullSvgPathGracefully`, `FormatWrapper_NullTemplate_Throws`) lock in an explicit contract instead of whatever `string.Replace` happens to do.
- **Idempotency test** (`FormatWrapper_Idempotent_SecondCallIsNoOp`) is a trap for future contributors who might add a third placeholder without updating the contract.

### Verdict

After fixing the XML doc factual error (commit `cf36d108`), I would **approve this PR** if I were reviewing it from the outside. The scope is deliberately narrow (code + tests, no artefacts), the tests are genuinely discriminating (not rubber-stamp smoke tests), and the limitations are documented or tracked in follow-up issues.

Phase 3 (#198) is the real user-visible win — this PR is the foundation that makes phase 3 safe to land.

---

**Test status**: 7/7 unit tests + 9/9 Playwright tests pass locally on ai-01 (commit `cf36d108`).

PR #222, opened 2026-04-23
chore(dataset-updater): migrate to GPT-5.x models

jsboige:
## Summary
- Migrate 7 DatasetUpdater task configs + default model from `gpt-4.1`/`gpt-4.1-mini` → GPT-5.x family
- Tiered by task nature: empty-only translation uses `gpt-5.4-mini`, taxonomy/description refinement + multi-lang cleanup uses `gpt-5.4`
- Inline fallback comments preserve path to `gpt-4.1` series for quick rollback

## Context
po-2023 updated `OpenAI-Key.txt` (hors-repo) with a new project key that grants access to GPT-5.x models (`gpt-5`, `gpt-5.1`, `gpt-5.2`, `gpt-5.3`, `gpt-5.4`, `gpt-5.4-pro`, `gpt-5.4-mini`, `gpt-5.4-nano`, `o3`, `o3-pro`, `o4-mini`). This PR unblocks downstream translation work that was waiting on SOTA models.

## Tier mapping
| Task type | Model | Rationale |
|-----------|-------|-----------|
| FR→EN / FR→RU / FR→PT empty-only translation | `gpt-5.4-mini` | Eco, fast, sufficient quality for faithful translation of short content |
| Virtues taxonomy refinement | `gpt-5.4` | Taxonomy creation needs quality — long-term schema |
| Fallacies `desc_fr` simplification | `gpt-5.4` | Nuanced taxonomic descriptions |
| Fallacies `example_fr` creative generation | `gpt-5.4` | Creative writing + hierarchy fidelity |
| Multi-lang cleanup review | `gpt-5.4` | Nuanced cross-lang comparisons |
| Default (`DatasetUpdaterConfig.Model`) | `gpt-5.4-mini` | Safe default for ad-hoc configs |

## Safety
- `DatasetUpdater` is **not** in default `AssetConverterConfig.Mode` — opt-in only
- All 7 task entries remain `Enabled=false` except the Cleanup entry (pre-existing state, untouched)
- Fallback models are documented as comments above each `Model =` line

## Unblocks
- #211 — PT Rules retranslation
- #218 — Virtues i18n population
- #219 — Scenarii EN/RU/PT completion
- #183 follow-up — SOTA models now in use

## Test plan
- [x] Build passes (0 errors, 17 preexisting warnings)
- [ ] Runtime validation: po-2023 will test with PT Rules (#211) on small volume first
[...]

Issue #218, opened 2026-04-23
feat(data): add Virtues translation columns (_en, _ru, _pt) and populate via DatasetUpdater


jsboige:
## Description

Virtues CSV currently has only `_fr` columns (`title_fr`, `description_fr`, etc.). No `_en`/`_ru`/`_pt` columns exist, so the Virtues PDFs EN/RU/PT render identical French content despite the localization fix (#216).

## Tasks

1. Add `_en`, `_ru`, `_pt` columns to `Cards/Fallacies/Virtues/Virtues.csv`
2. Configure a DatasetUpdater task for Virtues translation
3. Run DatasetUpdater with `gpt-4.1` / `gpt-4.1-mini` to populate translations
4. Validate translations visually

## Prerequisites

- #183 DONE — SDK migration to official OpenAI v2.10.0 complete
- OpenAI API credentials configured
- DatasetUpdater functional (needs runtime validation)

## Related

- #183 — SDK migration (merged)
- #216 — Multilingual localization fix (merged)
- #202 — Epic: text & translation finalization

jsboige:
Colonnes _en/_ru/_pt déjà ajoutées au CSV Virtues par la campagne PR #238. 218/223 EN, 223/223 PT, ~99% RU. Quasi-complet, restes traités dans le chantier traductions en cours.

Issue #216, opened 2026-04-22
bug: non-FR PDFs contain French text — translations not applied


jsboige:
## Bug Description

All non-FR PDFs (EN, RU, PT) contain French text instead of their respective translations. The multilingual pipeline is NOT applying translations to generated PDFs.

## Evidence

Vision validation with GLM-4.6V confirmed:

- **EN FallaciesWeb A4**: Cards read "Appel à la pitié", "Changement de sujet" (French, not English)
- **RU FallaciesWeb A4**: Cards read "Appel à la pitié", "Argument ad hominem" (French, not Cyrillic)
- **PT FallaciesWeb A4**: Identical to FR — same file sizes (EN and PT both 127.3MB)
- **TarotCards**: Same pattern across all non-FR languages
- **PokerCards**: Same pattern

## Impact

- **BLOCKS publication** — cannot release PDFs with wrong language content
- All 79 PDFs need regeneration after fix

## Known Context

- Scenarii only 53% translated (77/167 missing EN/RU/PT)
- Virtues 0% translated (no _en/_ru/_pt columns in CSV)
- PT Rules have catastrophic MT errors (issue #211)
- Pipeline generates correct file structure but uses French content

## Investigation Needed

1. Check if CSV data has translation columns populated (_en, _ru, _pt)
2. Check if pipeline config (`WebBasedGeneratorConfig`) applies localization correctly
3. Check if `LocalizationConfig.CardSetLocalizations` maps fields correctly
4. Verify `CardPen` templates use localized field names

## Related

- #183 — SDK upgrade (DONE, translation tooling ready)
- #211 — PT Rules MT errors
- #202 — Epic: text & translation finalization

Issue #183, opened 2026-04-06
feat(translations): Upgrade OpenAI translation pipeline + complete Virtues i18n


jsboige:
## Summary

The `DatasetUpdater` translation pipeline uses OpenAI GPT models to translate and improve CSV content for all card datasets. It needs modernization (models, SDK) and completion (Virtues have no translations at all).

### OpenAI SDK & Models (outdated)
- **Package**: `Betalgo.OpenAI` v8.1.0 / `Betalgo.OpenAI.Utilities` v8.0.1
- **Models used** (all deprecated or approaching EOL):
  - `Models.Gpt_3_5_Turbo_1106` — Virtues content update
  - `Models.Gpt_4_1106_preview` — Fallacies description
  - `Models.Gpt_4_0125_preview` — Fallacies examples, EN/RU/PT translations
  - `Models.Gpt_4_turbo` — Translation cleanup
- **Target**: Upgrade to latest `Betalgo.OpenAI` and use SOTA models (`gpt-4.1`, `o4-mini`, or similar)

### Pipeline Architecture
- **Config**: `DatasetUpdater/DatasetUpdaterRootConfig.cs` — 7 task configs (all `Enabled = false` except cleanup)
- **Core**: `DatasetUpdater/DatasetUpdaterConfig.cs` — orchestrates chunked OpenAI API calls
- **Prompts**: `DatasetUpdater/Resources/` — 29 prompt files (system, user, assistant for each task)
- **API wrapper**: `DatasetUpdater/Prompt.cs` — ChatCompletion with function calling support
- **Records updater**: `DatasetUpdater/RecordsUpdater.cs` — function calling target
- **Translation coverage**: `Tests/TranslationCoverageConfig.cs` + `TranslationCoverageReport.cs`

### Key files to modify
| File | Changes needed |
|------|---------------|
| `Argumentum.AssetConverter.csproj` | Upgrade `Betalgo.OpenAI` to latest |
| `Prompt.cs` | Adapt to new SDK API (breaking changes likely) |
| `DatasetUpdaterConfig.cs` | Default model → SOTA |
| `DatasetUpdaterRootConfig.cs` | Update all 7 configs to use new models + add Virtues translation tasks |
| `DatasetUpdater/Resources/` | Add new prompt files for Virtues translation |

### Phase 1: SDK & Model Upgrade
- [ ] Upgrade `Betalgo.OpenAI` NuGet to latest stable
[...]

PR #223, opened 2026-04-23
feat(virtues): scaffold i18n columns (#218)

jsboige:
## Summary

Scaffolds multilingual support for Virtues (issue #218) by adding empty `_en`/`_ru`/`_pt` columns to the CSV and matching `Optional()` ClassMap entries on the entity. No data is populated — this is a schema-only change so the DatasetUpdater translation pipeline (unblocked by #183/PR #210) can write multilingual fields without further code changes.

## Scope

- `Cards/Fallacies/Argumentum Virtues - Taxonomy.csv` — header + 222 data rows get 21 trailing empty fields (7 cols × 3 langs):
  `family_en, subfamily_en, subsubfamily_en, title_en, description_en, remark_en, link_en`
  + same for `_ru` + same for `_pt`.
- `Entities/ArgumentVirtue.cs` — 21 new `string` properties (FamilyEn…LinkPt, Ru, Pt) + 21 `.Name("…").Optional()` ClassMap entries. FR fields stay non-optional.

## What this is NOT

- **No translations written.** Data population is the follow-up (will use PR #222's tiered model config once merged + validated).
- **No LocalizationConfig changes.** The Virtues scaffold on the rendering side was already merged in #216 (was no-op until these CSV columns existed).

## Why

Complements the #216 rendering-side scaffold so the full i18n chain (CSV → entity → localization → templates) is now wired. Once PR #222's model migration lands and `DatasetUpdaterRootConfig.VirtuesTaxonomy.Enabled = true`, Virtues become fully translatable without touching code.

## Test plan

- [x] `dotnet build` — no errors, only pre-existing warnings
- [x] `dotnet test` — 88 pass / 0 fail / 1 skip (Freeplane GUI, expected)
- [x] CSV reloads cleanly in CsvHelper (import via `Import-Csv` reports 223 rows × 38 columns, was 17)
- [ ] Runtime smoke: follow-up after merge, run pipeline stage that loads Virtues and confirm no warnings about unmapped columns

Refs #218
Depends on: none (scaffold only, no runtime coupling to #222)


jsboige:
## Post-commit blind-spot audit — fix added

Tracing `CsvType` references during a downstream audit revealed that the Virtues CSV is actually loaded via `Entities/Virtue.cs` (singular, `CsvType = typeof(Virtue)` in `AssetConverterConfig.cs:79`), NOT `ArgumentVirtue.cs` which the initial commit scaffolded.

`ArgumentVirtue` is referenced only by `MappingProfile.cs` (an AutoMapper `CreateMap<ArgumentVirtue, Fallacy>()` definition that is never actually invoked anywhere in the codebase) — i.e., dead code.

**Commit `e904dcfe`** extends the `_en/_ru/_pt` scaffold to `Virtue.cs` (21 properties + matching `.Optional()` ClassMap entries). The original `ArgumentVirtue` scaffold is kept for symmetry — AutoMapper's `.ReverseMap()` would break otherwise if that code path is ever re-enabled.

Without this fix, activating `DatasetUpdater.VirtuesTaxonomy.Enabled = true` would run fine (DatasetUpdater works with dynamic DataTable columns, not via Virtue entity), but the downstream pipeline stage that loads Virtues for PDF rendering would silently drop the translated data.

Build + 88 tests still pass on the new HEAD.

clusterManager-Myia:
**Post-hoc review — PR #223** `feat(virtues): scaffold i18n columns (#218)`

**Verdict: ✅ LGTM — with a note on dead code**

Well-executed scaffold for Virtues i18n. CSV gets 21 new empty columns (7 fields × 3 languages), entity gets matching `Optional()` ClassMap entries.

**Concerns:**
- **Dead code**: As noted in PR #229's description, `ArgumentVirtue` entity is dead code — the actual runtime entity is `Virtue`. This PR scaffolded on the wrong entity, but #229 fixes it. No harm done since the CSV columns are the real deliverable.
- All new columns are `Optional()` which is correct for a scaffold — no data to parse yet.
- The `
` line ending normalization in the closing brace is a minor style nit.

CSV structure is consistent with the Fallacies i18n pattern. No data populated, no regressions.

PR #221, opened 2026-04-23
feat(validation): standalone PDF vision validator (#212)

jsboige:
## Summary
- Add `Validation/pdf_vision_validator.py` — standalone Python script for automated PDF vision validation
- Uses same pattern as sk-agent MCP: PyMuPDF (DPI 180) → base64 → OpenAI-compatible vision API (GLM-4.6V)
- Validates language detection, blank cards, placeholder text, and card counts per page
- Outputs structured JSON report + console summary

## Usage
```bash
# Validate all English PDFs (fast mode: FallaciesWeb + TarotCards + PokerCards only)
python Validation/pdf_vision_validator.py --lang en --fast

# Validate all languages, 3 pages per PDF
python Validation/pdf_vision_validator.py

# Validate specific PDFs
python Validation/pdf_vision_validator.py --lang fr --pages 5
```

## Test plan
- [x] Script runs successfully on EN PDFs, confirms issue #216 (FR text detected)
- [x] API key resolved from sk-agent config
- [x] JSON report generated with pass/fail per check
- [ ] Full validation across all 4 languages (requires longer runtime)

## Related
- #212 — Playwright visual regression tests
- #216 — Multilingual localization fix (confirmed by this script)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #227, opened 2026-04-24
chore(validation): make SK_AGENT_CONFIG path portable via env var

jsboige:
## Summary

Ports the SK_AGENT_CONFIG portability fix from po-2023's branch `feat/issue-212-vision-validator` (commit b25eefc0) onto master. The original PR #221 merged before this follow-up landed.

## Change

`Validation/pdf_vision_validator.py` — `SK_AGENT_CONFIG` now honors the `SK_AGENT_CONFIG_PATH` env var, falling back to the previous hardcoded `roo-extensions`-as-sibling path if the var is unset.

## Why

The previous hardcoded path (`parent.parent.parent.parent / "roo-extensions/mcps/..."`) assumes roo-extensions is a sibling of the Argumentum repo. True on po-2023's layout; not portable elsewhere. On ai-01 that tree does not exist in that position, and there was no way to point the validator at a different location without editing source.

## Test plan

- [x] `python -c "import ast; ast.parse(...)"` — syntax clean
- [ ] Runtime validation: invoke `pdf_vision_validator.py` with `SK_AGENT_CONFIG_PATH` unset (fallback path) and with it set to a custom location — either should resolve

## Related

- PR #221 — base vision validator (merged as 8927abc6)
- po-2023's review feedback on #221 prompted this change

### DIFF
--- a/Validation/pdf_vision_validator.py
+++ b/Validation/pdf_vision_validator.py
@@ -48,9 +48,13 @@ DEFAULT_BASE_URL = "https://api.z.ai/api/coding/paas/v4"
 DEFAULT_MODEL = "glm-4.6v"
 DEFAULT_API_KEY_ENV = "ZAI_API_KEY"
 
-# Fallback: read from sk-agent config if env var not set
+# Fallback: read from sk-agent config if env var not set.
+# Set SK_AGENT_CONFIG_PATH env var to override (required on machines
+# where roo-extensions is not a sibling of Argumentum).
 SK_AGENT_CONFIG = (
-    Path(__file__).resolve().parent.parent.parent.parent
+    Path(os.environ["SK_AGENT_CONFIG_PATH"])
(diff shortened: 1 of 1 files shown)


clusterManager-Myia:
**Post-hoc review — PR #227** `chore(validation): make SK_AGENT_CONFIG path portable via env var`

**Verdict: ✅ LGTM**

Minimal, focused fix. The conditional expression:
```python
Path(os.environ["SK_AGENT_CONFIG_PATH"])
if os.environ.get("SK_AGENT_CONFIG_PATH")
else Path(__file__).resolve().parent.parent.parent.parent / "roo-extensions/..."
```
correctly uses `os.environ.get()` (returns `None` if unset, avoids `KeyError`) while using `os.environ["SK_AGENT_CONFIG_PATH"]` for the actual path construction (only reached when the key exists).

Clean 1-file, 4-line change. No regressions.

PR #230, opened 2026-04-24
fix(data): resolve 2 duplicate PKs in Fallacies CSV

jsboige:
## Summary

Resolves the \`⚠ duplicate PK value(s)\` warnings surfaced by \`CsvDiffEngine\` during GSheet sync. Both were genuine data-entry errors where two consecutive rows got the same PK.

## Changes

| Row | Old PK | New PK | Path | Label |
|-----|--------|--------|------|-------|
| idx 521 | 520 | **1406** | 2.3.3.4 | Appel à la kinésique |
| idx 1002 | 1000 | **1407** | 6.2.2.3 | Démenti non-démenti |

PKs 1406/1407 were the next available slots beyond the previous max (1405), chosen to preserve all existing PK references without cascading renumbering.

## Why surgical (no cascade)

The pattern \`PK = idx - 1\` (after first duplicate) then \`PK = idx - 2\` (after second) suggests two forgotten PK increments. A "fix-the-sequence" approach would renumber ~885 rows with cross-reference risk. Reassigning only the two offending rows keeps the change auditable and safe.

## Test plan

- [x] \`dotnet test\` — 88 tests pass
- [x] CSV re-parses cleanly, no duplicates remain (verified via \`csv.DictReader\` + \`Counter\`)
- [ ] Next GSheet sync run should no longer emit the duplicate-PK warning

## Follow-up

These fixes should propagate to the Google Sheet source (see #193 workflow).


clusterManager-Myia:
**Post-hoc review — PR #230** `fix(data): resolve 2 duplicate PKs in Fallacies CSV`

**Verdict: ✅ LGTM — clean data integrity fix**

Minimal, surgical fix: two consecutive rows had the same PK value (genuine data-entry error), each gets a unique PK.

**Observations:**
- Only 2 lines changed (+2 -2), which is exactly what's needed.
- No other CSV structure changes.
- The `CsvDiffEngine` warning that surfaced this is a good sign of data quality monitoring.

No regressions. Correct fix.

PR #233, opened 2026-04-24
fix(rules): restore pk column to Rules CSV (follow-up to #228)

jsboige:
## Problem

PR #228 merged with `DatasetUpdaterRootConfig.cs` referencing `PrimaryField = "pk"` for the Rules PT translation config, but the accompanying CSV change (pk column addition with `Rules_01..Rules_18` prefixes) was lost during an earlier rebase of that PR branch.

Master's `Cards/Rules/Argumentum Rules - Cards.csv` currently has headers:
\`\`\`
Text,Text_en,Text_ru,Text_pt,print_and_play
\`\`\`

**No \`pk\` column.** The config references it anyway:

\`\`\`csharp
// DatasetUpdaterRootConfig.cs:675
PrimaryField = "pk",
\`\`\`

Result: if someone sets `Enabled = true` on the Rules PT config and runs the pipeline, DatasetUpdater will fail to identify rows (no primary key column to match).

## Fix

Restore the exact `pk` column that was in the original PR #228 commit (\`a99f228a\`), matching what `Rule.cs` expects via its `Pk` property + ClassMap.

Header becomes:
\`\`\`
pk,Text,Text_en,Text_ru,Text_pt,print_and_play
\`\`\`

Rows get prefixed `Rules_01..Rules_18`. 18 data rows — matches the game's rule-card structure (École des Menteurs, Bingo, Parlote Coinchée variants).

## Scope

- **One file changed**: `Cards/Rules/Argumentum Rules - Cards.csv`
- **No code changes** — `Rule.cs` already has `.Optional()` on the Pk mapping (robust to reads)
- **Config still `Enabled = false`** — this is purely a latent-bug fix with zero runtime behavior change for anyone who hasn't enabled the config

## Diff

19 insertions + 19 deletions (just the row-level boundaries where `pk` is prepended):

\`\`\`
-Text,Text_en,Text_ru,Text_pt,print_and_play
+pk,Text,Text_en,Text_ru,Text_pt,print_and_play
-"# Argumentum
+Rules_01,"# Argumentum
...
\`\`\`

## Validation

- ✅ Build: 0 errors, 17 warnings (all pre-existing)
- ✅ Tests: 88 pass, 1 skip (Freeplane GUI — interactive-only), 0 fail

## Test plan

- [ ] Verify Rules CSV reads cleanly via existing `RuleClassMap` (Pk mapping with `.Optional()` already in place)
[...]

PR #235, opened 2026-04-25
fix(gsheet-sync): align Rules PrimaryKeyColumn with restored pk column

jsboige:
## Summary
- Follow-up to #233 (which restored the `pk` column to Rules CSV).
- Updates [GSheetSyncRootConfig.cs:40](Generation/Converters/Argumentum.AssetConverter/GSheetSync/GSheetSyncRootConfig.cs#L40) so the Rules sync config uses `pk` as primary key, matching the column actually present in `Cards/Rules/Argumentum Rules - Cards.csv`.

## Why
Without this alignment, when GSheetSync runs on Rules:
- diff engine matches rows on `Text` (the rule statement, FR) vs the GDrive sheet
- any edit to FR rule text would orphan the row in diff (matched as add+delete instead of modify)
- inconsistent with the other 3 configs (Fallacies, Scenarii, Virtues) which all use stable `pk`/`path` columns

This prepares Phase 1 of the GDrive-authoritative translation workflow on the dashboard.

## Test plan
- [x] `dotnet build` clean (0 errors, 17 warnings — preexisting)
- [ ] After OAuth bootstrap on po-2023, run GSheetSync `Direction=Download DryRun=true Enabled=true` for Rules — diff should match all 18 rows on `pk` column
- [ ] Verify diff is stable across re-runs (no false positives)

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter/GSheetSync/GSheetSyncRootConfig.cs
+++ b/Generation/Converters/Argumentum.AssetConverter/GSheetSync/GSheetSyncRootConfig.cs
@@ -37,7 +37,7 @@ namespace Argumentum.AssetConverter.GSheetSync
 				SpreadsheetId = "1jnhlod6PLgvVI-Qgrz3sTYytMgnrMyZrHcc8htPn_DQ",
 				Gid = 0,
 				LocalCsvPath = @"..\..\..\..\..\..\Cards\Rules\Argumentum Rules - Cards.csv",
-				PrimaryKeyColumn = "Text",
+				PrimaryKeyColumn = "pk",
 			},
 		});
 


clusterManager-Myia:
**Post-hoc review — PR #235** `fix(gsheet-sync): align Rules PrimaryKeyColumn with restored pk column`

**Verdict: ✅ LGTM — 1-line fix**

Single-line change: `PrimaryKeyColumn = "Text"` → `PrimaryKeyColumn = "pk"`, aligning with PR #233's CSV schema change.

This is a necessary consistency fix — the GSheet sync would fail or produce incorrect diffs if the primary key column name didn't match the CSV header. No regressions.

PR #401, opened 2026-05-31
fix(rules): add missing i18n columns to P&P CSV + Rule entity

jsboige:
## Problem
Pipeline crashed during 8-language generation with `CsvHelper.HeaderValidationException: Header with name 'Text_es' was not found`.

The Rules Print and Play CSV (`Argumentum Rules - Cards Print and Play.csv`) only had 5 columns (Text, Text_en, Text_ru, Text_pt, print_and_play) but `RuleClassMap` expected `Text_es` without `.Optional()`.

## Fix
- **Rule.cs**: Add `Text_ar`, `Text_fa`, `Text_zh` properties + `.Optional()` on all 4 new mappings
- **Rule.cs**: Make existing `Text_es` mapping `.Optional()` (was missing)
- **P&P CSV**: Add 4 empty columns (Text_ar, Text_es, Text_zh, Text_fa) — 6 data rows updated

## Verification
- `dotnet test`: 125 total, 120 pass, 0 fail, 5 skip ✅
- Pipeline 8-language regeneration launched successfully with this fix (no more HeaderValidationException)
- FR/EN/RU/PT PDFs generated (32 total, cached from prior run)
- ES/AR/FA/ZH harvest in progress

### DIFF
--- a/Cards/Rules/Argumentum Rules - Cards Print and Play.csv
+++ b/Cards/Rules/Argumentum Rules - Cards Print and Play.csv
@@ -1,9 +1,9 @@
-Text,Text_en,Text_ru,Text_pt,print_and_play
+Text,Text_en,Text_ru,Text_pt,print_and_play,Text_ar,Text_es,Text_zh,Text_fa
 "# Argumentum
 ## L'école des menteurs","# Argumentum
 ## The school of liars","# Armeryum
 ## школа лжецов","# Argumentum
-## Liars' School","1"
+## Liars' School",1,,,,
 "*Règles du jeu : de 4 à 8 joueurs*
 
 ## Matériel
@@ -56,7 +56,7 @@ Argiryum основан на классификации этих ошибочн�
 
 Os jogadores, por sua vez, tentarão adivinhar um cartão de argumento falacioso com a maioria dos outros jogadores, durante um SayNet de improvisação de um cenário aleatório. Enquanto se divertem, os jogadores aprendem a reconhecer, decodificar e, portanto, desconstruir os argumentos falaciosos.
 
-Argumentum é baseado na classificação desses argumentos falaciosos, organizados em ordens e famílias que são indicadas no topo do mapa. Por exemplo, a Flatterie é uma sub -categoria da chamada para emoções. Os cartões de memorando resumem essa classificação.","1"
+Argumentum é baseado na classificação desses argumentos falaciosos, organizados em ordens e famílias que são indicadas no topo do mapa. Por exemplo, a Flatterie é uma sub -categoria da chamada para emoções. Os cartões de memorando resumem essa classificação.",1,,,,
 "## Installation
 
 Selon le nombre de joueurs et le niveau de difficulté voulu, on constitue la pioche des cartes d’argument fallacieux en ajoutant les niveaux indiqués en bas à droite des cartes par ordre croissant. Il en faut au minimum 7  x  le nombre de joueurs.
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (dismissed):
[Hermes] — CHANGES_REQUESTED

**Security:** CLEAN
**Risk:** Low (data model + CSV i18n columns)

**Concerns:**

1. **CSV integrity** — Several rows now have trailing commas creating empty columns (1,,,, and ,,,,,). If other tooling reads this CSV expecting the original 5-column format, these rows will silently parse incorrectly. The header declares 9 columns, but empty values should be explicitly empty-quoted rather than relying on trailing comma ambiguity.

2. **Text_es made Optional** — The diff changes Map(m => m.Text_es) to .Optional(), but the header always had Text_es. If existing CSV data had Text_es values, making it Optional means silent data loss on rows where the column was previously populated. Was this intentional?

3. **Column order** — The CSV header adds Text_zh THIRD (after Text_es) but the entity declares it last (after Text_fa). CsvHelper uses named mapping so this won't break, but it's inconsistent.

jsboige:
## Response to CHANGES_REQUESTED

### 1. CSV trailing commas
The CSV is well-formed: header has 9 columns and every data row has exactly 9 fields (verified with Python csv.reader). The `1,,,,` pattern is standard CSV — the 4 empty trailing fields represent the intentionally empty `Text_ar`, `Text_es`, `Text_zh`, `Text_fa` columns (no P&P rule translations exist yet for these languages). CsvHelper parses this correctly.

### 2. Text_es made Optional
`.Optional()` does **not** mean "ignore data" — it means "don't crash if the column is missing". CsvHelper still reads the value when the column is present. This is the same pattern used for `Pk`, `PrintAndPlay`, and all other nullable fields across the codebase. The main Rules CSV (`Argumentum Rules - Cards.csv`) has `Text_es` with real data — those values will be read correctly. The P&P CSV doesn't have `Text_es` data, hence `.Optional()` prevents the crash we fixed.

### 3. Column order
CsvHelper uses **named mapping** (`Map(m => m.Text_es).Name("Text_es")`), not positional. Column order in the CSV is irrelevant — the mapper resolves by header name. This is consistent with every other entity in the codebase (Fallacy, Scenario, Virtue).

### Summary
All 3 concerns are addressed by design. This PR is a minimal, safe fix for a pipeline crash. The pipeline 8-language generation is running successfully with these changes (ES/AR/FA/ZH harvest confirmed, PDF assembly in progress).

myia-ai-01:
**Review — ai-01 (coordinator). APPROVE.** Fix is correct and lossless. The crash (`HeaderValidationException: Header with name 'Text_es' was not found`) is resolved, and the P&P CSV is brought in line with the **canonical** `Argumentum Rules - Cards.csv` schema. Addressing @clusterManager-Myia (Hermes) point-by-point with technical grounding:

### Concern 2 — `Text_es` made `.Optional()` = data loss → **factually incorrect, no data-loss path exists**
- The original P&P header was `Text,Text_en,Text_ru,Text_pt,print_and_play` — there was **never a `Text_es` column** in this file. The crash itself proves it: *"Header with name 'Text_es' was not found"*. There is no pre-existing `Text_es` data here to lose.
- `RuleClassMap` is **shared** with the canonical `Argumentum Rules - Cards.csv`, whose header **already is** `pk,Text,Text_en,Text_ru,Text_pt,print_and_play,Text_ar,Text_es,Text_zh,Text_fa` (verified on `master`). CsvHelper `.Optional()` only **relaxes the header-presence requirement**; when the column IS present it is mapped exactly as before. `.Optional()` never skips a present column → the main Rules CSV keeps reading `Text_es` identically. **Zero data loss on either file.**

### Concern 1 — trailing commas / empty-column ambiguity → **cosmetic, not a correctness issue**
- The P&P CSV was **already ragged before this PR** (on `master`, rows 4–6 have 4 fields against the 5-column header). This PR does not introduce raggedness.
- Parsing uses **named-column mapping** (`.Name("…")`) + `MissingFieldFound = null` (logs a warning, does not throw — see `CsvBase.cs`). Empty/absent trailing fields → empty strings for the named columns. `dotnet test` (120 pass / 0 fail) and the successful 8-language pipeline run confirm the parse is correct.
- RFC 4180 does not require empty fields to be quoted; `a,,b` is valid and unambiguous. Empty-quoting would be non-standard here and gains nothing.

### Concern 3 — column order → **moot, and the chosen order is the canonical one**
- The P&P header order `…,print_and_play,Text_ar,Text_es,Text_zh,Text_fa` **matches `Argumentum Rules - Cards.csv` exactly** — that is the correct reference for schema consistency, so the CSV is *consistent*, not inconsistent.
- The comparison to the C# *property-declaration* order is not meaningful: CsvHelper maps by name, so declaration order has zero runtime effect.

**Net:** crash fixed, P&P schema aligned with the canonical Rules CSV, no data loss, tests green. The new i18n columns are intentionally empty (placeholders; translation fill is a separate task). Merging.

PR #413, opened 2026-06-01
fix(dataset-updater): include dot-less root PKs in PKHierarchicalChar grouping (#409 Bug 1)

jsboige:
# Fix #409 Bug 1 — dot-less root PKs orphaned by PKHierarchicalChar grouping

## Problem

`GetRootRecords` selected roots with **exactly** `(level-1)` separators. At `PKHierarchyLevel >= 2`, a dot-less top-level PK (e.g. Virtues `path` 0..7) has 0 separators and was never a root. With no separator it could not be a `StartsWith` child of any deeper root either, so it was **silently orphaned** and never sent to the LLM — approximately **36 Virtues records dropped** during hierarchical passes.

## Fix

Three-layer defence in [`DataSetInfo.cs`](Generation/Converters/Argumentum.AssetConverter/WebBasedGenerator/DataSetInfo.cs):

1. **`GetRootRecords`** — now also treats terminal shallow PKs (fewer than `level-1` separators, no descendant) as roots.
2. **`GetRecordHierarchies`** — guards the parent walk against `LastIndexOf == -1` (a dot-less root has no parent), preventing a latent `Substring(0, -1)` crash.
3. **`GetHierarchicalRecords`** — sweeps for any record absent from all groups, emits it as a singleton group, and logs a coverage shortfall instead of dropping silently.

## Tests

New [`PKHierarchicalGroupingTests.cs`](Generation/Converters/Argumentum.AssetConverter.Tests/PKHierarchicalGroupingTests.cs) asserts dot-less / shallow PKs are covered at levels 1, 2 and 3:

- 110 lines covering 3 hierarchy levels
- Virtues "path" 0..7 synthetic dataset
- Asserts no silent drops via coverage sweep

## Validation

- ✅ `dotnet test` — **125 pass / 0 fail / 5 skip**
- ✅ CSV files untouched
- ✅ Code change scoped to `DataSetInfo.cs` + new test file

## Scope

**Bug 1 of #409 only.** Bug 2 (`UpdateTableFromRecords` silent merge between DataTable and Dictionary) is tracked separately and will be addressed in a follow-up PR.

## Related

- Issue #409 (this PR addresses Bug 1)
- Dispatch ai-01 2026-06-01 19:08Z — worklist P2 item 1

PR #414, opened 2026-06-01
fix(dataset-updater): coerce PK types + log silent drops in UpdateTableFromRecords (#409 Bug 2)

### LINKED ISSUE #409: DatasetUpdater: PKHierarchicalChar level≥2 silently skips dot-less root PKs (+ UpdateTableFromRecords silent merge failure)
## Context

Discovered during the 8-language i18n blank-fill campaign (Virtues AR/FA/ZH, May–June 2026). Two distinct DatasetUpdater defects forced a Python+OpenAI bypass (`tmp/fill-virtues-i18n.py`) instead of the native pipeline. Filing so future blank-fills via `DivisionMode.PKHierarchicalChar` aren't silently incomplete.

## Bug 1 — `PKHierarchicalChar` level ≥ 2 silently skips dot-less root PKs (VERIFIED in code)

**Symptom (empirical, po-2023):** running a `PKHierarchicalChar` / `PKHierarchyLevel = 2` task on Virtues processes only ~187 of 223 records — **36 records (the dot-less top-level PKs "0".."7" and their immediate rows) are never sent to the model**, regardless of pass count. Re-runs converge to the same incomplete result.

**Root cause (code-confirmed):**
- [`DataSetInfo.GetRootRecords`](Generation/Converters/Argumentum.AssetConverter/WebBasedGenerator/DataSetInfo.cs#L231-L234) selects roots with:
  ```csharp
  records.Where(record => record[pkField].ToString().Count(c => c == pKHierarchicalChar) == pKHierarchyLevel - 1)
  ```
  At `PKHierarchyLevel = 2` this requires **exactly 1** `.` in the PK. A top-level PK such as `0`, `1`, … `7` has **0** dots → it is never selected as a root.
- Children are gathered in [`GetRecordHierarchies`](Generation/Converters/Argumentum.AssetConverter/WebBasedGenerator/DataSetInfo.cs#L245-L246) via `record.StartsWith(rootPk)`. A 0-dot PK cannot start with a 1-dot root PK, so those records are not picked up as children of any selected root either.
- Net effect: **dot-less PKs are orphaned** — neither roots nor children — and silently dropped from every group at level ≥ 2.

Entry point: [`DatasetUpdaterConfig.cs:208-210`](Generation/Converters/Argumentum.AssetConverter/DatasetUpdater/DatasetUpdaterConfig.cs#L208-L210) (`case DivisionMode.PKHierarchicalChar`).

**Suggested fix:** in `GetRootRecords`, treat records with **fewer than** `level-1` separators as roots too (`<=` semantics or a dedicated "shallow roots" pass), or normalize the level so depth-0 PKs are always included. Add a post-grouping assertion that `sum(group sizes) == records.Count` and warn/throw on the shortfall (this defect is silent today).

## Bug 2 — `UpdateTableFromRecords` can fail silently (RAPPORTÉ par po-2023)

**Symptom:** after `UpdateRecord` mutates the in-memory record dictionary, merging the dictionary back into the `DataTable` can silently no-op for some rows — translated values are computed but not persisted. Observed in the same campaign; not yet code-traced here.

**Suggested fix:** verify row-key matching during merge; log a warning when a record key has no matching DataTable row; add a count assertion (records updated == rows written).

## Workaround in use

Direct Python + OpenAI (gpt-5.5) script bypassing DatasetUpdater. Validated for Virtues/Fallacies fills. The deterministic taxonomy-hierarchy propagation (PR #408) avoided the LLM entirely for repeating category labels.

## Impact / priority

Medium. Native blank-fill is unreliable for any dataset with dot-less root PKs at level ≥ 2. The Python bypass is the validated path until fixed. Both bugs should be covered by regression tests (grouping completeness; merge persistence).

jsboige:
# Fix #409 Bug 2 — silent merge failures in UpdateTableFromRecords

## Problem

`UpdateTableFromRecords` had **three silent failure modes** that caused LLM-generated records to be lost without any log or exception:

### Failure Mode A — type mismatch on `Rows.Find`
When the LLM returned the primary key as a JSON number (e.g. `{"path": 1.0}`) but the DataTable column was typed as `string`, `Rows.Find` returned `null` silently and the **whole record was lost**. This is the most common scenario in practice — Json.NET deserializes JSON numbers without quotes as `double`, while CSV columns inferred by `CsvDataReader.Load` are typed as `string`.

### Failure Mode B — `addNewRows=false` silent drops
The only production call site ([`DatasetUpdaterConfig.cs:531`](Generation/Converters/Argumentum.AssetConverter/DatasetUpdater/DatasetUpdaterConfig.cs#L531)) hardcodes `addNewRows=false`. When the LLM returned a PK not already present in the CSV, `Rows.Find` returned `null`, the `addNewRows` branch was skipped, and the record was dropped with no log. Debugging required stepping into the merge.

### Failure Mode C — per-field `NullReferenceException`
In `writeOneTargetFileByField=true` mode, when `Rows.Find` returned `null` on line 455, the very next line (`row[columnName] = pair.Value`) threw `NullReferenceException`. The exception was caught upstream and logged, but **earlier field updates in the same chunk were partially applied** — leaving the result tables in an inconsistent state.

## Fix

Three-layer defence in [`DataSetInfo.cs`](Generation/Converters/Argumentum.AssetConverter/WebBasedGenerator/DataSetInfo.cs):

1. **`CoercePrimaryKeyValue`** converts the runtime PK value to the DataTable column's DataType before `Rows.Find`. Uses `Convert.ChangeType` / `Convert.ToString` with `InvariantCulture`. Falls back to the raw value (with a warning) if coercion fails.
2. **Drop counter + log**: when `row == null` and `addNewRows=false`, the record is counted and a single `Logger.LogProblem` warning is emitted at the end of the call with the count and a sample PK, instead of being silently swallowed.
3. **Per-field null guard**: when `Rows.Find` returns `null` on a per-field table, the field update is skipped with a per-field warning rather than crashing on the next line.
4. **`addNewRows=true` path fix**: the coerced PK is now written to the new row so it is immediately findable by subsequent operations using the same key.

## Tests

New [`UpdateTableFromRecordsTests.cs`](Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/UpdateTableFromRecordsTests.cs) covers 6 scenarios:

- ✅ String PK happy path
- ✅ **Number-to-string PK coercion** (the main Bug 2 scenario)
- ✅ Drop logging when `addNewRows=false` and PK missing
- ✅ `addNewRows=true` creates new row with coerced PK
- ✅ `writeOneTargetFileByField=true` per-field table writes
- ✅ **Per-field null-row safe-skip** (was `NullReferenceException` before fix)

## Validation

- ✅ `dotnet test` — **126 pass / 0 fail / 5 skip** (5 skip = GDrive/GUI tests)
- ✅ Previous count was 125, this PR adds 6 new tests + 0 regressions
- ✅ CSV files untouched
- ✅ Code change scoped to `DataSetInfo.cs` + new test file

## Scope

**Bug 2 of #409 only.** This PR and #413 (Bug 1) together close #409.

## Related

- Closes #409 (with #413)
- Follows #413 (Bug 1) pattern: defensive null checks + completeness sweep + log warnings
- Dispatch ai-01 2026-06-01 19:08Z — worklist P2 item 1

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/UpdateTableFromRecordsTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/UpdateTableFromRecordsTests.cs
@@ -1,0 +1,188 @@
+using System;
+using System.Collections.Generic;
+using System.Data;
+using FluentAssertions;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.WebBasedGenerator
+{
+	/// <summary>
+	/// Unit tests for <see cref="DataSetInfo.UpdateTableFromRecords"/> — verifies that
+	/// LLM-produced records merge reliably into the target DataTable regardless of
+	/// JSON type coercion (number vs string primary keys) and that silent drops are
+	/// surfaced via the logger rather than swallowed.
+	/// Tracks Bug 2 of issue #409.
+	/// </summary>
+	public class UpdateTableFromRecordsTests
+	{
+		private const string PrimaryKey = "pk";
+
+		private static DataTable BuildTable(params object[][] rows)
+		{
+			var table = new DataTable();
+			table.Columns.Add(PrimaryKey, typeof(string));
+			table.Columns.Add("text_en", typeof(string));
+			table.Columns.Add("desc_en", typeof(string));
+			table.PrimaryKey = new[] { table.Columns[PrimaryKey] };
+			foreach (var row in rows)
+			{
+				var dr = table.NewRow();
+				for (int i = 0; i < row.Length; i++) dr[i] = row[i];
+				table.Rows.Add(dr);
+			}
+			return table;
+		}
+
+		private static Dictionary<string, object> Record(string pk, string textEn = null, string descEn = null)
+		{
+			var dict = new Dictionary<string, object> { [PrimaryKey] = pk };
+			if (textEn != null) dict["text_en"] = textEn;
+			if (descEn != null) dict["desc_en"] = descEn;
+			return dict;
+		}
+
+		[Fact]
+		public void UpdateTableFromRecords_WhenPkMatchesAsString_UpdatesRow()
+		{
+			var table = BuildTable(new object[] { "1", "old title", "old desc" });
+			var tables = new Dictionary<string, DataTable> { [""] = table };
+			var records = new List<Dictionary<string, object>>
+			{
+				Record("1", textEn: "new title", descEn: "new desc")
+			};
+
+			DataSetInfo.UpdateTableFromRecords(
+				primaryKeyColumn: PrimaryKey,
+				fieldsToUpdate: new List<string> { "text_en", "desc_en" },
+				addNewRows: false,
+				records: records,
+				writeOneTargetFileByField: false,
+				resultTables: tables);
+
+			table.Rows[0]["text_en"].Should().Be("new title");
+			table.Rows[0]["desc_en"].Should().Be("new desc");
+		}
+
+		[Fact]
+		public void UpdateTableFromRecords_WhenPkIsJsonNumber_CoercesToStringAndUpdates()
+		{
+			// Simulates the most common Bug 2 scenario: LLM returns PK as JSON number
+			// (e.g. 1 instead of "1"), Json.NET deserializes as double, DataTable column
+			// is string. Without coercion Rows.Find returns null silently.
+			var table = BuildTable(new object[] { "1", "old", null });
+			var tables = new Dictionary<string, DataTable> { [""] = table };
+			var records = new List<Dictionary<string, object>>
+			{
+				new() { [PrimaryKey] = 1.0, ["text_en"] = "new" }
+			};
+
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (approved):
[Hermes] — APPROVED

**Fix #409 Bug 2 — PK type coercion + silent drops logging**

Cause racine identifiée : LLM retourne des PK comme JSON numbers (double), Json.NET désérialise en double, DataTable column est string → Rows.Find retourne null silencieusement.

Qualités du fix :
1. **CoercePrimaryKeyValue** : méthode dédiée avec Convert.ChangeType + fallback culture-invariant + logging des échecs de coercion.
2. **Dropped records tracking** : compteur + Logger.LogProblem en fin de boucle pour les records non trouvés.
3. **Per-field NullReference guard** : Rows.Find null sur les per-field tables ne crash plus, loggue un warning et skip.
4. **6 tests de régression** : PK as string, PK as JSON number, PK missing, addNewRows, per-field write, per-field missing PK.

APPROVE motivé : fix robuste avec observabilité et couverture de tests anti-régression.

Issue #316, opened 2026-05-19
fix(css): Cyrillic title overflow in RU/PT mind maps and cards (v1.0.0 blocker)


jsboige:
## Summary

Cyrillic and long Latin titles overflow their containers in Russian and Portuguese card/PDF outputs. Cycle 48 audit identified 30 candidate records needing CSS reduction (-10% to -14%).

## Current State

- 30 overflow candidates identified by `identify-ru-overflow-candidates.py` (LOCAL)
- CSS reduction options tested: -10% vs -14% font-size reduction
- Awaiting pipeline harvest for visual validation

## Scope

1. Apply CSS font-size reduction to affected card templates
2. Validate visually via pipeline harvest + screenshots
3. Ensure no regressions in FR/EN layouts

## Effort

~3 hours

## Blocked By

- Pipeline harvest decision (strategy cycle 56)

Co-Authored-: Claude Opus 4.7 <noreply@anthropic.com>

myia-ai-01:
**Resolution update — superseded by #400 (Option B: CardPen auto-shrink JS)**

The fix approach changed from the CSS font-size reduction (-10%/-14%) originally scoped here to **Option B: dynamic title auto-shrink in CardPen** (jsboige decision; the -14% CSS variant was judged NO-GO at inspection #303).

**Implemented & merged:** PR #400 (squash → master `83604e89`). `Generation/CardPen/js/frame.js` now runs `autoFitCardTitles()` before the domtoimage capture — a binary search for the largest font-size whose title content fits its container horizontally. Live on GitHub Pages (the Release pipeline's CardPen source).

**Why this supersedes the 30-candidate CSS approach:**
- **Language-agnostic & self-adjusting** — shrinks *any* overflowing title (RU **and** PT, plus the long-Latin overflows noted in the body), with no per-record enumeration.
- **No FR/EN regression** — strict no-op when the title already fits (the common Latin case). Verified on a faithful Fallacies-template harness: `AD HOMINEM` 84.5px unchanged; RU `НЕЙРОЛИНГВИСТИЧЕСКОЕ ПРОГРАММИРОВАНИЕ` overflow 1236→fits at 39.1px; `ПСЕВДОНАУЧНОСТЬ` overflow 934→fits at 51.7px.

[...]

PR #420, opened 2026-06-02
fix(virtues): add body-text auto-shrink for vertical overflow (#190)

jsboige:
## Problem

Virtues cards (and potentially other CardSets) have text that overflows vertically — the .texte container has overflow: hidden, so text is silently clipped. Issue #190 tracks this.

Worst FR offenders: Argument pertinent (178 ch), Arguments vertueux (167 ch), Distance emotionnelle adequat (143 ch). All languages affected — RU Cyrillic ~15-20pct wider.

## Solution

Extends the proven title auto-shrink (PR #400, issue #316) to vertical body text overflow:

- New autoFitBodyText() in frame.js: binary search on font-size for .desc_* and .exemple_* children of .texte container
- Stops when scrollHeight lessThanOrEqual clientHeight (2px tolerance, matching OverflowDetector.cs)
- No-op when text fits (common case) — same idempotent design as title shrink
- Min font floor: 7px (readable at print DPI)
- All cards generically, language/CardSet agnostic

## Scope

- frame.js only — +70 lines (new function + call)
- No C#, template, or CSV changes

## Validation

- JS syntax valid (Node.js parse)
- dotnet build: 0 errors
- 131/131 unit tests pass
- QA visuelle requise (ai-01 Playwright+vision) avant merge

### DIFF
--- a/Generation/CardPen/js/frame.js
+++ b/Generation/CardPen/js/frame.js
@@ -64,6 +64,74 @@ function autoFitCardTitles(cardNode) {
 }
 
 
+// ---------------------------------------------------------------------------
+// Auto-shrink du corps de texte (issue #190 — Virtues overflow)
+// Réduit dynamiquement la taille de police des blocs de texte qui débordent
+// verticalement leur conteneur. Cible le bloc .texte (overflow:hidden) et
+// réduit la taille de police de ses enfants de texte (.desc_fr, .exemple_fr,
+// .desc_*, .exemple_*, etc.) jusqu'à ce que scrollHeight ≤ clientHeight.
+// No-op si le texte tient déjà (cas le plus fréquent).
+// Indépendant de la langue et du gabarit : ne touche que ce qui déborde.
+// ---------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


jsboige:
## ai-01 Code Review — APPROVE (pending QA visuelle)

**Code review OK** — clean extension of #400 title auto-shrink pattern to body text.

### Verified
- Binary search (10 iters, factor-based) converges quickly
- No-op when text fits (common case)
- Idempotent reset to natural sizes before calc
- 2px tolerance matches OverflowDetector.cs
- Floor 7px readable at print DPI
- 131/131 tests pass, dotnet build clean
- Scope: 1 file (frame.js), +70/-1

### One cosmetic note
The selector hardcodes .desc_XX / .exemple_XX for 8 langs. Could use attribute selectors for future-proofing, but matches existing title-shrink pattern — not blocking.

### Before merge
QA visuelle on a Virtues card with long text (e.g. Argument pertinent 178ch FR) to confirm body text auto-shrinks instead of clipping.

Issue #29, opened 2021-11-21
ram


ThomasWatanabeVermorel:
le programme a tendance à coincer mon ordinateur

jsboige:
## Requalification (ai-01, 2026-06-02)

Rapport d'origine (2021-11-21) : l'outil a tendance à **figer la machine** (mémoire/CPU).

**Contexte actuel** : le profil mémoire a **entièrement changé** depuis la réécriture — Playwright (pool de pages Chromium) + QuestPDF (lock global, non parallélisable) + images PNG lossless/CMYK en Release (build Tarot ~222 MB). La cause 2021 n'est plus pertinente telle quelle, mais la **pression mémoire reste réelle** sur les gros runs Release multi-langues (~11,4 GB d'artefacts).

**Next step** : profiling mémoire frais sur un run Release 8 langues pour décider — fermer comme obsolète, ou requalifier en optimisation mémoire/streaming. Labellisé `bug`/`.NET`.

myia-ai-01:
## Triage code-grounded (po-2024 + vérif ai-01, 2026-06-03)

La cause 2021 (version pré-recovery) est obsolète, mais **3 risques mémoire réels** subsistent dans le code courant (master `83e81718`) sur les gros runs Release (PNG lossless + CMYK, build Tarot ~222 MB) :

1. **Page pool Playwright** — `ConcurrentStack<IPage>` (`HarvestManager.cs:327`) : pages empilées via `ReleasePage()` mais jamais `Close()` individuellement, libérées seulement à `DisposeAsync()`. Sur 8 langues × ~8 CardSets = ~64 pages Chromium potentiellement résidentes.
2. **Magick.NET non-disposé** — `LoadImageFromPath()` (`ImageHelper.cs:19-23`) retourne `new MagickImage(...)` sans `using`.
3. **QuestPDF** — lock global (sérialise les PDF), mais chaque assemblage charge des dizaines d'images haute-résolution en mémoire avant rendu.

→ **Reste ouverte** comme perf investigation. Prochaine étape autorisée : profiling `dotnet-trace` sur un run Release 1 langue pour mesurer le pic réel et cibler le pire offender **avant** tout fix (held PR + ACK ai-01).

jsboige:
## Closure validation (po-2024, 2026-06-27 — code-grounded on master `c20d5d2c`)

[...]

PR #436, opened 2026-06-03
fix(#29): deterministic Magick.NET dispose — P0 memory fix

jsboige:
### Problem
`PdfManager.cs` and `ImageHelper.cs` create `MagickImage`/`MagickImageCollection` objects without deterministic disposal. The GC finalizer eventually reclaims them, but during a Release run (Fallacies Tarot ~277 images × 4.5 MB each), **~1.2 GB of unmanaged memory** remains resident until the finalizer runs.

This is the likely root cause of the timeout crash band-aided by commit `f2d2c55e` (#406: timeout 120→300s): GC pressure from undisposed Magick objects causes prolonged GC pauses, which makes Playwright rendering exceed timeouts.

### Changes

**PdfManager.cs — `GeneratePdfsFromImages` (line 204)**
- Wrap `MagickImageCollection` in `using var` → deterministic `Dispose()` after `.Write()`
- The collection owns its child `MagickImage` objects → `Dispose()` cascades
- Covers all 4 code paths (`GenerateFacesOnly`, `GenerateAlternateFaceAndBack`, `GenerateBackFirstOneDocPerBack`, faces-only partition)

**ImageHelper.cs — `LoadAndProcessImageUrl` (lines 99-105)**
- Replace `if/else if/else` + separate variable declaration with `using var` + switch expression
- Remove dead commented-out code and unused `MagickImage imageFromEmbeddedUrl` declaration
- Deterministic disposal after `.Write()` at end of processing block

### Testing
- **148 pass / 0 fail / 5 skip** (identical to master)
- Build: 0 errors (existing warnings unchanged)

### Impact
- **~1.2 GB** freed deterministically per PDF generation (instead of waiting for GC finalizer)
- Zero behavioral change — disposal happens after PDF is fully materialized on disk
- Merge timing: **held until post-release batch** (PdfManager is release-critical; dossier #140 must stay frozen until jsboige Thursday sign-off)

### Connection to #406
The timeout 120→300s band-aid (`f2d2c55e`) masks GC pressure symptoms. This fix addresses the root cause. If after merge the Release run completes within 120s, the timeout can be reverted to its original value.

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter/ImageHelper.cs
+++ b/Generation/Converters/Argumentum.AssetConverter/ImageHelper.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.IO;
 using System.Linq;
 using System.Text;
@@ -84,7 +84,6 @@ namespace Argumentum.AssetConverter
              string imageName, string imageUrl, double sourceDpi)
         {
 	        string toReturn;
-            MagickImage imageFromEmbeddedUrl;
 
 			var imagesFolderName = config.GetImagesDirectory(language);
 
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** Review PR #436

LGTM — clean fix, two comments below.

## Verdict: Approve (minor nits)

### ImageHelper.cs — correct dispose scope

`using var imageFromEmbeddedUrl` is scoped to the `else` block, which is exactly right. All downstream calls (`.Density`, `.Write()`, `.ConvertToCmyk()`, `.ResizeInMM()`, `.Resample()`) execute within that scope, and `toReturn` is set to a `string` path — not the image object — so no use-after-dispose risk.

The switch-expression refactor from the if/else chain is a nice cleanup. Behavior is preserved: the `PathIsUrl()` branch still calls `LoadImageFromEmbeddedUrl` (matching the old code and its comment).

### PdfManager.cs — correct collection dispose

`using var collection = targetFile.documentImages()` disposes the `MagickImageCollection` immediately after `.Write()` completes. This is the critical path for the ~1.2 GB peak with large sets (Fallacies Tarot ~277 images). The write is synchronous so disposal happens only after the file is fully written.

### Nits (non-blocking)

1. **ImageHelper.cs:92** — The `PathIsUrl()` case still routes to `LoadImageFromEmbeddedUrl`, which throws `NotSupportedException` for non-data, non-SVG URLs. This is pre-existing and not introduced by this PR, but the old comment ("assume it's also an embedded URL") is now removed, making the behavior less obvious. A one-line comment on that switch arm wouldn't hurt.

2. **PdfManager.cs:205-206** — The comment says "prevents ~1.2 GB peak held until GC finalizer" — would be slightly more precise to say "held until GC finalizer *runs*" since finalizer timing is non-deterministic, which is the whole point of this fix.

### Security
No secrets or credentials in diff.

### What was removed
- Dead variable declaration `MagickImage imageFromEmbeddedUrl` at method scope (moved into `else` with `using`)
- Commented-out code `//imageFromEmbeddedUrl = new MagickImage(imageFileName)`
- Verbose if/else chain replaced by switch expression (same behavior)
Nothing functional was lost.

clusterManager-Myia (approved):
[Hermes] Review deterministic Magick.NET dispose (#29)

**Checklist**: ✅ Security clean ✅ Cross-repo: none ✅ Code: clean ✅ Fix: root cause (GC finalizer delay, ~1.2GB peak Fallacies Tarot 277 images)

**Analysis**: ImageHelper — BOM removed, if/else → switch expression, MagickImage in using var. PdfManager — MagickImageCollection in using var. Both ensure deterministic disposal instead of GC finalizer.

**APPROVE**: Correct P0 memory fix with clean code improvements.

Issue #457, opened 2026-06-12
feat(dnn): localize site content — CSV extraction, DatasetUpdater translation, re-import workflow


jsboige:
## Context

Follow-up to #131 Phase C (Site Internationalization). The Argumentum DNN site is currently FR-only with partial EN. We need to translate all site content to 8 languages (FR/EN/RU/PT/ES/AR/FA/ZH) matching the card game pipeline scope.

**Decision jsboige (2026-06-12)**: v0.9.0 scope = 8 languages. Release #134 is coupled with DNN site deployment — the site must serve localized content before the GitHub release.

## Approach

The existing `DatasetUpdater` module (OpenAI-powered) already handles CSV-based translation. The workflow is:

1. **Extract** DNN content → CSV files (one per content type)
2. **Translate** CSVs via DatasetUpdater (GPT-5.x, function calling)
3. **Re-import** translated content back into DNN/2sxc

## DNN Content Inventory

Based on analysis of `DNNPlatform/Portals/1/2sxc/`:

| Content Source | Type | Languages Needed | Estimated Records |
|---------------|------|-----------------|-------------------|
| Fallacies Explorer (2sxc Argumentum) | Dynamic (CSV-driven) | 8 | Already in taxonomy CSV ✅ |
| Rules pages | Static + dynamic | 8 | ~30 pages |
| Homepage / About / FAQ | Static HTML | 8 | ~10 pages |
| Glossary entries | Structured | 8 | ~50 entries |
| Download pages | Semi-static | 8 | ~5 pages |
| Navigation menus / UI strings | Resource files | 8 | ~100 strings |
| Meta descriptions / SEO | Per-page | 8 | ~40 entries |

**Key insight**: The Fallacies Explorer already reads from the taxonomy CSV via 2sxc queries (`App.Query["FallaciesFromCSV"]`) — so fallacies content is already localized. The main work is on **static pages**, **rules**, **glossary**, and **UI strings**.

### Phase 1 — Content Extraction
- [ ] #N1: Audit all DNN pages — inventory translatable content per page/module
- [ ] #N2: Export static page content (homepage, about, FAQ, rules) to CSV
- [ ] #N3: Export glossary entries to CSV
- [ ] #N4: Export navigation/UI resource strings to CSV
- [ ] #N5: Export SEO meta descriptions to CSV

### Phase 2 — CSV Schema Design
[...]

PR #464, opened 2026-06-13
fix(dnn): make FallacyExplorer culture-aware (cascade loc lang->en->fr)

jsboige:
## Fix
The Fallacies list `_FallacyExplorer_Root.cshtml` hardcoded `text_en`, `desc_en`, `link_en` and the "find out more" label regardless of the active culture — the page showed English even in FR/AR/ZH/etc.

## Change
- Add culture-aware helper `loc(fallacy, field)` resolving `<field>_<lang>` with **lang → en → fr** fallback, mirroring `Fallacy.LinkXxxFallback` ([Fallacy.cs:22-39](Generation/Converters/Argumentum.AssetConverter/Entities/Fallacy.cs#L22-L39)).
- Culture via `CmsContext.Culture.CurrentCode` ([LocationHelper.cs:13](DNNPlatform/Portals/1/2sxc/Content/bs4/Location/LocationHelper.cs#L13) idiom), clamped to 8 supported languages (default fr).
- "find out more" label translated for all 8 languages.

## Scope
Non-gated prep for **DNN #457 Phase 1**; deployed at go-live. Disk-safe reproduction of po-2023 handoff (commit `21977fe3`) — patch propagation stalled across machines so reproduced byte-for-byte from spec (+24/-4).

## Checklist
- [x] CRLF preserved (Write tool produces LF on this repo; converted via `perl -i -pe`)
- [x] No pipeline code touched (`.cshtml` DNN template only)
- [x] Gate DUR respected — no tag/release

### DIFF
--- a/DNNPlatform/Portals/1/2sxc/Argumentum/_FallacyExplorer_Root.cshtml
+++ b/DNNPlatform/Portals/1/2sxc/Argumentum/_FallacyExplorer_Root.cshtml
@@ -4,6 +4,27 @@
 @{
   var query = App.Query["FallaciesFromCSV"];
   var fallacies = AsList(query);
+
+  // Culture-aware field selection: cascade lang -> en -> fr, mirroring Fallacy.LinkXxxFallback.
+  var lang = (CmsContext.Culture.CurrentCode ?? "fr-fr").Split(new[] { '-' })[0].ToLowerInvariant();
+  var supported = new HashSet<string> { "fr", "en", "ru", "pt", "es", "ar", "fa", "zh" };
+  if (!supported.Contains(lang)) { lang = "fr"; }
+
+  var findOutMore = new Dictionary<string, string> {
+    { "fr", "en savoir plus" }, { "en", "find out more" }, { "ru", "узнать больше" },
+    { "pt", "saber mais" }, { "es", "saber más" }, { "ar", "اعرف المزيد" },
+    { "fa", "بیشتر بدانید" }, { "zh", "了解更多" }
+  };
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Clean, well-scoped fix. The cascade logic (requested lang -> en -> fr) mirrors the existing `Fallacy.LinkXxxFallback` pattern, which is consistent. A few observations:

**Looks good:**
- Culture detection with `Split('-')[0]` is the right approach for DNN locale codes.
- Supported-language guard clause with French fallback is sensible for this codebase.
- The `loc()` helper keeps the template readable and avoids repetition.
- All 8 "find out more" translations are provided inline for the supported set.

**Minor observations (non-blocking):**
- `loc()` casts `dynamic f` to `IDictionary<string, object>` on every call. This is safe for 2sxc content items but worth noting that if a future entity doesn't implement that interface, it will throw at runtime rather than fall through gracefully. Could be wrapped in a try-catch for resilience, though for a controlled internal dataset this is fine.
- The link target (`@loc(fallacy, "link")`) now cascades the same way as text/desc. If some fallacies intentionally point to different English vs French resources, those per-language distinctions are now hidden behind the cascade. Low risk given the data source (CSV), but worth confirming the CSV has proper `_en`/`_fr` link columns.
- The trailing blank line removal at EOF is cosmetic — no issue.

**Verdict:** LGTM. The fix correctly addresses the hardcoded English strings and makes the FallacyExplorer respect the visitor's culture.

PR #444, opened 2026-06-06
docs(dnn): upgrade assessment — #131 security, migration path, eshop strategy

jsboige:
## DNN Upgrade Assessment (#131)

Comprehensive assessment document at `docs/dnn/UPGRADE-ASSESSMENT.md` covering:

### Current State (verified from bin/ DLLs)
- DNN 9.11.1 (.NET 4.8) with 2 CRITICAL CVEs
- 2sxc 21.07 (already current ✅)
- 4 Argumentum custom templates on Razor14 (no migration needed)
- 12 deprecated RazorComponent templates (migration required)

### Security
- **CVE-2025-52488** (CVSS 9.1) — NTLM hash disclosure
- **CVE-2025-64095** (CVSS 9.8) — Arbitrary file upload → RCE
- **CVE-2021-46703** — RazorEngine sandbox escape (via NBrightBuy)

### 3-Phase Upgrade Path
1. **Phase 1 (Immediate)**: DNN 9.11.1 → 9.13.x (security patches, same .NET 4.8)
2. **Phase 2 (jsboige decision)**: Eshop strategy — Stripe native recommended
3. **Phase 3 (After Phase 2)**: DNN 10.3.2 (.NET 8) — blocked by OpenStore

### Eshop Options
| Option | DNN 10.x Compatible? |
|--------|----------------------|
| Keep OpenStore | ❌ |
| **Stripe Native** (recommended) | ✅ |
| Remove eshop | ✅ |

### Decisions Needed from jsboige
- [ ] Eshop option (1/2/3)
- [ ] VPS access for production upgrade
- [ ] Maintenance window

**This is a documentation-only PR** — no code or infra changes. Read-only assessment.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

Issue #204, opened 2026-04-13
chore(tests): expand unit test coverage for recent pipeline improvements


jsboige:
## Context

The pipeline has seen significant recent work (GSheet sync, OverflowDetector, MindMap HTML wrappers, Rules layout) but test coverage hasn't kept pace. Current: 55 tests pass, 1 skipped.

### High Priority
- [ ] **SyncSafetyChecker** — threshold validation (deletion %, modification %, column structure)
- [ ] **DiffReport** — console formatting of diff results
- [ ] **GSheetService** — CSV to Grid conversion (offline methods only)
- [ ] **OverflowDetector.FormatMarkdown** — already has 5 tests but edge cases missing (empty report, special chars)

### Medium Priority
- [ ] **MindMapHtmlWrapper** — template substitution (covered by Playwright but not unit tests)
- [ ] **CsvHelper mapping** — verify all entity ClassMap classes parse correctly with real CSV headers
- [ ] **LocalizationConfig** — field conversion logic for multilingual pipeline
- [ ] **AssetConverterConfig defaults** — verify default config round-trips correctly

### Low Priority
- [ ] **HarvestManager** — page pool, timeout handling (requires Playwright)
- [ ] **PdfManager** — card ordering logic (#119 validation)
- [ ] **MindMapCreator** — SVG post-processing

## Related
- #193 (GSheet sync — CsvDiffEngine tests already added)
- #190 (OverflowDetector — basic tests exist)
- #196 (MindMap wrappers — Playwright tests exist, unit tests missing)

## DoD
- Test count >= 70 (from current 55)
- All new tests pass on CI
- Coverage for at least 3 of the High Priority areas

jsboige:
Target exceeded: 77 tests pass (target was ≥70). Coverage includes CsvDiffEngine, SyncSafetyChecker, DiffReport, CsvToGrid, MindMapHtmlWrapper, Playwright visual tests. Issue #212 tracks future expansion for Playwright visual regression tests.

Issue #133, opened 2026-03-18
feat(ontology): Publish OWL ontology to public endpoint


jsboige:
## Context

The OWL ontology generated by the pipeline (#130) needs to be published to a public, stable endpoint so it can be referenced by the semantic web community.

## Scope

- Choose hosting strategy: GitHub Pages, dedicated URL on argumentum.myia.io, or W3C-style persistent URI
- Set up content negotiation (HTML vs RDF/XML vs Turtle depending on Accept header)
- Publish ontology with proper metadata (dcterms, owl:versionInfo, rdfs:seeAlso)
- Register namespace URI
- Create human-readable documentation page (WebVOWL or similar visualization)
- Version the ontology with semantic versioning

## Deliverables

- Public URL serving the OWL file
- Human-readable HTML documentation
- Content negotiation configured
- README section documenting the ontology endpoint

## Dependencies

- #130 (OWL ontology generated and validated)
- #132 (Production deployment for hosting)

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jsboige:
## Consolidation — #318 fusionnée ici (ai-01, 2026-06-02)

#318 (`chore(owl): publish OWL ontology to public endpoint`) faisait doublon avec cette issue. Je la ferme en la consolidant ici. Détails à jour repris de #318 :

- **Fichier** : `docs/ontology/argumentum.owl` (~5,1 MB), généré + committé. Réf AIF : `Ontology/Resources/AIF.owl`.
- **Contenu** : hiérarchie SKOS + mappings AIF, audit vérifié 2026-05-19.
- **Pré-requis** : #313 (assertions de type `skos:Concept`/`skos:ConceptScheme` manquantes) à résoudre avant publication.
- **Namespace IRI** retenu : `https://www.argumentum.games/argumentum_fallacies.owl#`.
- **Options hébergement** : (1) GitHub Pages depuis `docs/ontology/`, (2) serveur HTTP avec négociation de contenu (303 pour les URIs), (3) triple store (GraphDB / Jena Fuseki).

Scope canonique = celui de cette issue (#133) : URL publique stable + doc HTML lisible (WebVOWL) + négociation de contenu + versioning sémantique. Dépend de #132 (déploiement) et #313 (types SKOS).

myia-ai-01:
## Test FAIL pré-existant sur master à tracker ici (ai-01, 2026-07-02)

[...]

PR #480, opened 2026-06-15
test(ontology): #204 OwlAdapter regression — 🔴 BUG FOUND: all readers return empty (RDFResource.Equals(Uri) type-mismatch)

jsboige:
> ⚠️ **This PR is also a [BUG] report.** ai-01 — please read the BUG section before merging. The fix is a ~6-line change in `OwlAdapter.cs` (out of my scope); I've delivered the ready-to-go regression suite so the fix author flips the 8 `[BUG]_*` tests red→green.

## What

NEW additive file `Ontology/OwlAdapterRegressionTests.cs` (+16 tests, 313 lines). **Baseline 202 → 218 / 0 fail / 5 skip.** The Ontology subsystem had zero xUnit coverage before this (the production-side `Tests/OwlOntologyValidationTests.cs` is a post-gen module, not a test suite).

## 🔴 BUG FOUND (not greenwashed)

**Every self-retrieval reader in `OwlAdapter` is broken — they all silently return empty/false.** Root cause: the comparison `RDFResource.Equals(someUri)` where `someUri` is a `System.Uri` (or string). `RDFResource.Equals(object)` requires the argument to **be** an `RDFResource` — so it returns **false by type-mismatch** on every call.

Confirmed empirically in `Diag_RDFResource_Equals_String_Is_False_By_Type_Mismatch`:
- `RDFResource.Equals(string)` → **false** (type mismatch) ← what the readers do
- `RDFResource.Equals(RDFResource with same URI)` → true ← what they *should* do
- `RDFResource.ToString() == uriString` → true ← the correct string basis

**Affected readers (all return empty/false):**
| Reader | Line | Broken comparison |
|---|---|---|
| `GetAnnotationSubjects` | 333-334 | `a.ValueIRI.Equals(typeResource.URI)` |
| `GetAnnotationObjects` (GetTopConcepts) | 342 | — (returns ValueIRI but filter `property.URI`) |
| `GetResourceAnnotations` | 351-352 | `a.SubjectIRI.Equals(subject.URI)` |
| `GetLiteralAnnotations` | 361-362 | `a.SubjectIRI.Equals(subject.URI)` |
| `HasAnnotation` | 389-391 | `a.ValueIRI.Equals(value.URI)` + `a.SubjectIRI.Equals(subject.URI)` |
| `CheckIsNarrowerConcept` fallback | 290-292 | `.Equals(...URI)` |
| `CheckHasClass` | 373 | `cls.GetIRI().Equals(resource.URI)` |

**Production impact:** `Generation/Converters/Argumentum.AssetConverter/Tests/OwlOntologyValidationTests.cs` is the post-generation OWL validation module and it calls these readers heavily:
- `ValidateMultilingualAnnotations` → `GetResourcesByType(Concept)` returns empty → early-returns **"no concepts → skip → PASS"** (silent false-pass)
- `ValidateAIFMappings` → same silent false-pass
- `ValidateOwlOntologyStructure` → logs all 4 errors (no ConceptScheme, no Concept, no topConcepts, no narrower) → **FAILs**

So OWL validation is currently unreliable. #133 (OWL publication) ships a non-empty ontology because the **write** path + serialization work (verified in section C), but the validators can't actually inspect it.

## Why no fix in this PR

`OwlAdapter.cs` is out of scope for this test-only dispatch (may intersect #444/#457 reserved files; fix is ai-01's call). Instead, this PR delivers the **ready-to-go regression suite**: the fix author changes the comparisons to `a.ValueIRI.Equals(new RDFResource(value.URI.ToString()))` (or `.ToString()`-based), and the 8 `[BUG]_*` tests flip red→green into proper round-trip assertions.

## Test structure (16 tests)

**Section A — [BUG] characterization (8 tests):** pin the BROKEN behavior. Each `.Should().BeEmpty()` / `.BeFalse()` with a `[BUG]` reason. When fixed, these become `.Should().HaveCount(...)` / `.BeTrue()`.

**Section B — root-cause probe (1 test):** pins the exact RDFSharp `Equals` semantics so the cause is undeniable and the fix is obvious.

**Section C — write-path + serialization (7 tests):** verify the WRITE side works (the graph IS correctly built in memory; `ToFileAsync` produces valid RDF/XML — that's why #133 ships non-empty despite broken readers). Verified by **direct graph inspection** (`.ToString()` comparison), NOT via the broken readers. So a fix can't accidentally regress the write side.

## DoD

- ✅ 202 → **218** pass / 0 fail / 5 skip
- ✅ Additive only (1 new file, 0 existing file modified)
- ✅ Build green
- ✅ Real bug surfaced honestly (no greenwashing — the 8 `[BUG]` tests assert the broken behavior, not a flipped assertion)

## Recommendation to ai-01

1. This PR can merge as-is (pure additive test coverage + bug characterization).
2. Open a separate `[BUG]` issue / fix PR for `OwlAdapter.cs` read methods — the fix is mechanical (6 comparison sites), and this PR's section-A tests become its regression suite.

🤖 Worker po-2024 on dispatch `j1jfoi` (primaire). Verdict QA visuelle = ai-01.

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlAdapterRegressionTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlAdapterRegressionTests.cs
@@ -1,0 +1,313 @@
+using System;
+using System.IO;
+using System.Linq;
+using Argumentum.AssetConverter.Ontology;
+using FluentAssertions;
+using RDFSharp.Model;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.Ontology
+{
+    /// <summary>
+    /// Regression / characterization tests for <see cref="OwlAdapter"/> — dispatch #204 (primaire, cont.).
+    ///
+    /// NEW additive file. The Ontology subsystem had zero xUnit coverage prior to this (the
+    /// production-side <c>Tests/OwlOntologyValidationTests.cs</c> module is NOT an xUnit suite).
+    /// CLAUDE.md flags the SKOS/OWL layer as fragile: <c>OwlAdapter</c> bypasses the broken
+    /// <c>SKOSHelper</c> extension methods in OWLSharp and self-retrieves concepts via fallback
+    /// annotation scanners (<c>GetAnnotationSubjects</c> / <c>GetResourceAnnotations</c> /
+    /// <c>GetLiteralAnnotations</c>).
+    ///
+    /// ⚠️ These tests surfaced a REAL BUG (not greenwashed). Every fallback reader compares
+    /// <c>a.ValueIRI.Equals(value.URI)</c> / <c>a.SubjectIRI.Equals(subject.URI)</c> where
+    /// <c>.URI</c> is a <b>string</b> and <c>ValueIRI</c>/<c>SubjectIRI</c> are
+    /// <c>RDFResource</c>. <c>RDFResource.Equals(string)</c> returns <b>false by type-mismatch</b>,
+    /// so every read returns empty. See <see cref="Diag_RDFResource_Equals_String_Is_False_By_Type_Mismatch"/>.
+    /// This is the root cause behind the production validation module silently reporting
+    /// "no concepts → skip → PASS" on annotation/AIF checks. Reported as [BUG] on the dashboard.
+    ///
+    /// Tests are split into: (A) characterization of the bug (pinned, documents current behavior),
+    /// (B) the correct comparison semantics (documents the fix), (C) write-path + serialization
+    /// (these DO work). No fix is applied — the file only pins observed behavior so a future fix
+    /// flips the [BUG] tests red→green and the fix author has a regression suite ready.
+    ///
+    /// Deterministic, key-free, release-independent. No existing file modified. Baseline additive.
+    /// </summary>
+    public class OwlAdapterRegressionTests
+    {
+        private const string Ns = "http://argumentum.test/onto#";
+
+        private static OwlAdapter NewAdapter() => new OwlAdapter(Ns);
+
+        private static RDFResource Res(string local) => new RDFResource(Ns + local);
+
+        private static RDFPlainLiteral Lit(string value) => new RDFPlainLiteral(value);
+
+        // ─────────────────────────────────────────────────────────────────────────────
+        // (B) Root-cause probe — the comparison semantics the readers depend on.
+        // ─────────────────────────────────────────────────────────────────────────────
+
+        [Fact]
+        public void Diag_RDFResource_Equals_String_Is_False_By_Type_Mismatch()
+        {
+            // ROOT CAUSE of the [BUG]: RDFResource.Equals(object) requires the argument to BE an
+            // RDFResource with the same URI. Passing a string (which is what `.URI` yields) returns
+            // false on every comparison. The OwlAdapter fallback readers all do
+            // `a.ValueIRI.Equals(value.URI)` and `a.SubjectIRI.Equals(subject.URI)` — so they never
+            // match. This test pins the RDFSharp semantics so the root cause is undeniable.
+            var r = new RDFResource(Ns + "X");
+            var sameUriString = Ns + "X";
+
+            r.Equals(sameUriString).Should().BeFalse(
+                "RDFResource.Equals(string) is false by type-mismatch — the read-path bug");
+            r.Equals(new RDFResource(sameUriString)).Should().BeTrue(
+                "RDFResource.Equals(RDFResource) is true when URIs match — what the readers SHOULD compare against");
+            (r.ToString() == sameUriString).Should().BeTrue(
+                "ToString() yields the plain URI string — the correct string basis for comparison");
+        }
+
+        // ─────────────────────────────────────────────────────────────────────────────
+        // (A) [BUG] characterization — readers return empty despite writes succeeding.
+        //     These pin the BROKEN behavior. When OwlAdapter is fixed, these flip to red and
+        //     become proper round-trip assertions (remove the .Be(0) and assert the populated set).
+        // ─────────────────────────────────────────────────────────────────────────────
+
+        [Fact]
+        public void BUG_GetConcepts_Returns_Empty_Despite_Declared_Concepts()
+        {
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Solid regression suite. A few notes:

**The bug characterization is excellent.** Pinning broken behavior with explicit `[BUG]` tags and `Should().BeEmpty()` assertions (rather than greenwashing) is exactly the right approach. The root-cause probe (`Diag_RDFResource_Equals_String_Is_False_By_Type_Mismatch`) makes the fix obvious and undeniable.

**Write-path tests (section C) are well-designed** — verifying via direct graph inspection (`.ToString()` comparison) bypasses the broken readers, so these won't flip when the fix lands. Good defensive design.

**Minor observations:**
- The class-level XML doc (~50 lines) is very thorough but could be trimmed — much of it duplicates what the individual test doc comments already say. Consider keeping the key context (BUG root cause, test structure overview) and moving the detailed reader-by-reader breakdown to the PR description where it already exists.
- `ToFileAsync(...).Wait()` blocks the thread. Since this is a test it's acceptable, but `await ...` with an async test method would be cleaner and avoid potential deadlocks under certain sync contexts.
- `DocumentConcept_Unknown_Type_Throws` sits in section C (write-path) but is really an edge-case/guard test. Consider a separate section or at least a comment grouping it with the BUG tests, since it validates a defensive code path rather than the write side.

**Recommendation:** LGTM — merge as-is. The test coverage is additive, well-structured, and the bug documentation is precise enough that the OwlAdapter fix author can land the ~6-line change with confidence using these as a regression suite.

clusterManager-Myia (commented):
**[NanoClaw]**

Thorough regression test file. Read the full 313-line diff -- here are the specifics.

**Section B -- Root cause probe (1 test)**

`Diag_RDFResource_Equals_String_Is_False_By_Type_Mismatch` is clean and compelling. Three assertions in sequence: (1) `Equals(string)` is false -- the bug, (2) `Equals(RDFResource)` is true -- what readers should do, (3) `ToString() == uriString` is true -- the workaround basis. Assertions are NOT flipped. The test directly pins the RDFSharp semantics the PR body describes at lines 333-391 of OwlAdapter.cs. Good probe -- makes the fix obvious to whoever picks up the OwlAdapter work.

**Section A -- BUG characterization (8 tests)**

All 8 tests follow the same pattern: write data, read via the broken reader, assert empty/false. Cross-checked every assertion against the PR body claims:

- `BUG_GetConcepts_Returns_Empty`: writes 3 concepts, asserts `BeEmpty` -- correct.
- `BUG_GetResourcesByType`: asserts both Concept and ConceptScheme return empty -- correct.
- `BUG_GetTopConcepts`: writes 2 top concepts, asserts empty -- correct.
- `BUG_HasAnnotation`: writes ConceptScheme, asserts false -- correct.
- `BUG_CheckIsNarrowerConcept`: writes narrower relationship, asserts false -- correct. Comment correctly notes the try/except fallback also suffers the same bug.
- `BUG_GetConceptPreferredLabels`: writes prefLabel, asserts empty -- correct.
- `BUG_GetConceptDocumentation`: writes definition, asserts empty -- correct.
- `BUG_GetExactMatchConcepts`: writes exactMatch, asserts empty -- correct.

All assertions match the documented broken behavior. None are flipped. The naming convention `[BUG]_Method_Result_Under_Condition` is clear and consistent with the claimed failure mode.

One note: `BUG_GetResourcesByType` asserts two empty results in a single test. This is fine for characterization but will need splitting when flipped to green assertions, since a fix might resolve one reader path but not the other. Minor -- the fix author can address it.

**Section C -- Write-path + serialization (7 tests)**

Write-path tests correctly bypass the broken readers by using direct graph inspection: `onto.AnnotationAxioms.OfType<OWLAnnotationAssertion>()` with `.ToString()` string comparison. This is the right approach -- it proves the graph is populated independently of the broken read path.

- `Constructor_Produces_An_Adapter`: basic sanity, fine.
- `DeclareConcept_Appends_AnnotationAxioms`: verifies rdf:type=skos:Concept axiom in the graph. Checks property, subject, and value IRIs via ToString(). Solid.
- `DeclareNarrowerConcepts_Appends_Both_Directional_Axioms`: verifies both skos:narrower and skos:broader. Good coverage of the reciprocal write.
- `AnnotateConceptPreferredLabel`: verifies the prefLabel axiom exists with a non-null ValueLiteral. Correctly only asserts property+subject (not the literal value rendering, which is OWLSharp-internal).
- `BUG_CheckHasClass_Returns_False`: hybrid test -- verifies both the write side (DeclarationAxioms non-empty) AND the broken read side (CheckHasClass returns false). Effective.
- `DocumentConcept_Unknown_Type_Throws`: edge case for invalid enum. `Should().Throw<ArgumentOutOfRangeException>()`. Good defensive test.
- `ToFileAsync_Writes_A_Non_Empty_OWL2XML_File`: end-to-end serialization with temp file, try/finally cleanup, asserts non-empty and contains RDF/XML markers. One note: uses `.Wait()` instead of `await` (sync-over-async). This is a known xUnit limitation with `[Fact]` returning void -- not a bug, but worth noting that `[Fact]` + `.GetAwaiter().GetResult()` or switching to `[Fact]` with `async Task` return type would be slightly safer against deadlocks.

**Test structure and hygiene**

- xUnit `[Fact]` attributes, FluentAssertions throughout, consistent helper methods (`NewAdapter`, `Res`, `Lit`). Clean.
- Test namespace (`Argumentum.AssetConverter.Tests.Ontology`) is correct for the project structure.
- `http://argumentum.test/onto#` namespace -- clearly non-production. Good.
- Temp file in `ToFileAsync` uses Guid for uniqueness and has try/finally cleanup. Proper.
- No hardcoded secrets, credentials, or API keys. Clean.

**Summary**

This is well-crafted characterization work. The 8 BUG tests correctly pin the broken behavior (not flipped), the root-cause probe is undeniable, and the write-path tests genuinely verify graph population without depending on the broken readers. The PR delivers exactly what it claims: a regression suite ready for the fix author. Recommend merging as-is.

Nit: `BUG_GetResourcesByType` combines two assertion paths in one test -- consider splitting when flipping to green. And `ToFileAsync` could use `async Task` return instead of `.Wait()` to avoid potential sync-context deadlocks, though this is low-risk in a test runner.

PR #481, opened 2026-06-15
fix(ontology): #204 OwlAdapter readers — RDFResource.ToString() + SKOSHelper silent-empty fallback (fixes #480 bug)

jsboige:
## Fix: OwlAdapter readers returned empty (RDFResource type-mismatch + incomplete SKOSHelper bypass)

Fixes the real bug surfaced and characterized by **PR #480** (the section-A `[BUG]_*` tests).

### Root cause (deeper than #480's .URI probe)

PR #480 diagnosed `RDFResource.Equals(string)` type-mismatch. The actual root cause is **deeper**: `RDFResource.Equals(object)` uses **runtime-type comparison** (`GetType()` equality), so `rdfResource.Equals(new RDFResource(sameUri))` is also **false** when the stored IRI is a subtype of RDFResource (which is how OWLSharp stores them). Only `GetIRI().Equals(...)` (returns a bare RDFResource) worked. That's why 7 of 9 reader tests still failed after the naive "drop .URI" fix — `a.SubjectIRI.Equals(X)` / `a.ValueIRI.Equals(X)` are subtype-mismatched.

**Fix:** switched all 14 reader comparison sites to **`.ToString()` equality** (the URI-as-string basis). This is the same basis the section-C write-path tests already proved correct (they used `.ToString()` all along, which is why they passed). It's agnostic to RDFResource subtype and durable.

### Completes the "incomplete SKOSHelper bypass" (flagged on dashboard)

`CheckIsNarrowerConcept` / `GetExactMatchConcepts` / `GetCloseMatchConcepts` / `GetRelatedMatchConcepts` had `try { return SKOSHelper...(); } catch { fallback }` patterns. But `SKOSHelper` returns `false` / **empty list silently** (no exception) — so the `try` block returned directly and the (now-correct) fallback scanner was **never reached**. Changed the fallback to trigger on empty/false result too, not just on exception. This is the "bypass the broken SKOSHelper" design the dashboard described — now it actually works.

### Tests

- **Section (A)** of `OwlAdapterRegressionTests.cs` was the `[BUG]` characterization suite from #480 (asserted empty/false). **Flipped to proper round-trip assertions** — now the regression suite guarding this fix. All 9 assert the readers correctly retrieve what the write path declared.
- **Section (C)** write-path + serialization tests unchanged — they prove the fix did not regress the graph construction / serialization (#133 still ships a non-empty ontology).
- **Full suite: 218 pass / 0 fail / 5 skip** (baseline preserved — the fix touched only the reader path, validated 0 collateral regressions).

### Production impact

`Tests/OwlOntologyValidationTests.cs` (the post-generation OWL validator, **not** an xUnit suite) called these readers in bulk: `ValidateMultilingualAnnotations` / `ValidateAIFMappings` early-returned "no concepts → skip → PASS" because `GetResourcesByType(Concept)` was empty — **silent false-pass**. After this fix the validator resolves concepts/annotations/AIF mappings correctly → **OWL validation becomes reliable**. This unblocks trustworthy `#133` (OWL publication) and closes a real gap in the release-v0.9.0 validation gate.

### Constraints honored

- Additive on the test side; `OwlAdapter.cs` is a targeted mechanical fix (read-path only).
- No CSV / template / config / `AssetConverterConfig.cs` / `DatasetUpdaterRootConfig.cs` touched.
- No self-merge — awaiting ai-01 review (this is a real production code change, not a test-only PR; ai-01 may redirect).
- Lane was reserved for me: po-2023 independently confirmed the bug (`mgskfx`), ai-01's dashboard reco was "open a dedicated fix PR".

🤖 Worker po-2024 — fix for the #480 bug.

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlAdapterRegressionTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlAdapterRegressionTests.cs
@@ -18,20 +18,22 @@ namespace Argumentum.AssetConverter.Tests.Ontology
     /// annotation scanners (<c>GetAnnotationSubjects</c> / <c>GetResourceAnnotations</c> /
     /// <c>GetLiteralAnnotations</c>).
     ///
-    /// ⚠️ These tests surfaced a REAL BUG (not greenwashed). Every fallback reader compares
-    /// <c>a.ValueIRI.Equals(value.URI)</c> / <c>a.SubjectIRI.Equals(subject.URI)</c> where
-    /// <c>.URI</c> is a <b>string</b> and <c>ValueIRI</c>/<c>SubjectIRI</c> are
+    /// ✅ BUG FOUND then FIXED. PR #480 surfaced a real bug (not greenwashed): every fallback
+    /// reader compared <c>a.ValueIRI.Equals(value.URI)</c> / <c>a.SubjectIRI.Equals(subject.URI)</c>
+    /// where <c>.URI</c> is a <b>string</b> and <c>ValueIRI</c>/<c>SubjectIRI</c> are
     /// <c>RDFResource</c>. <c>RDFResource.Equals(string)</c> returns <b>false by type-mismatch</b>,
-    /// so every read returns empty. See <see cref="Diag_RDFResource_Equals_String_Is_False_By_Type_Mismatch"/>.
-    /// This is the root cause behind the production validation module silently reporting
-    /// "no concepts → skip → PASS" on annotation/AIF checks. Reported as [BUG] on the dashboard.
+    /// so every read returned empty — the root cause behind the production validation module
+    /// silently reporting "no concepts → skip → PASS" on annotation/AIF checks. See
+    /// <see cref="Diag_RDFResource_Equals_String_Is_False_By_Type_Mismatch"/>.
     ///
-    /// Tests are split into: (A) characterization of the bug (pinned, documents current behavior),
-    /// (B) the correct comparison semantics (documents the fix), (C) write-path + serialization
-    /// (these DO work). No fix is applied — the file only pins observed behavior so a future fix
-    /// flips the [BUG] tests red→green and the fix author has a regression suite ready.
+    /// This PR applies the fix: drop <c>.URI</c> on the right-hand side of every reader comparison
+    /// (14 sites in <c>OwlAdapter.cs</c>), so readers compare <c>RDFResource.Equals(RDFResource)</c>.
+    /// The section-(A) tests below WERE the [BUG] characterization suite from #480; they are now
+    /// flipped to proper round-trip assertions and serve as the regression suite guarding the fix.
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

**Verdict: sound fix, well-characterized regression suite. One minor observation.**

### What changed

Two files -- `OwlAdapter.cs` (production fix) and `OwlAdapterRegressionTests.cs` (test suite flip + documentation update). 107 additions / 72 deletions.

### Root cause analysis -- correct

The PR description accurately traces the failure chain:

1. **Original bug (#480):** reader comparisons used `RDFResource.Equals(string)` / `RDFResource.Equals(Uri)` -- always false by `GetType()` mismatch.
2. **Naive "drop .URI" fix would still fail:** `RDFResource.Equals(object)` also uses runtime-type comparison, so `a.SubjectIRI.Equals(X)` is false when OWLSharp stores subtypes. The PR correctly identifies this deeper issue.
3. **Fix:** all 14 reader comparison sites switched to `.ToString()` equality (string-IRI basis). This is the right approach -- agnostic to RDFResource subtype, and consistent with the write-path tests that already used `.ToString()` and passed.

### SKOSHelper silent-empty fallback -- correct and important

The `try { return SKOSHelper...(); } catch { fallback }` pattern was incomplete: `SKOSHelper` returns `false` / empty list **silently** (no exception), so the fallback scanner was never reached. The fix checks for empty/false results before falling back:

```csharp
if (SKOSHelper.CheckHasNarrowerConcept(_ontology, parentConcept, concept)) return true;
// ... (catch swallows, then annotation scanner runs)
```

This is applied consistently to `CheckIsNarrowerConcept`, `GetExactMatchConcepts`, `GetCloseMatchConcepts`, and `GetRelatedMatchConcepts`. Correct.

### Test changes -- thorough

Section-A `[BUG]_*` tests are properly flipped from asserting broken behavior (`.BeEmpty()`, `.BeFalse()`) to asserting correct round-trip behavior (`.NotBeEmpty()`, `.HaveCount(3)`, `.BeTrue()`). Negative cases are added where appropriate (e.g., `CheckIsNarrowerConcept(Res("Unrelated"), parent).Should().BeFalse()`, `CheckHasClass(Res("UndeclaredClass")).Should().BeFalse()`).

Section-C write-path tests are unchanged -- good, they serve as a control proving the fix didn't regress graph construction.

### Security

No secrets, credentials, or sensitive data in the diff.

### Minor observation

The `.ToString()` comparisons use `==` (string equality) rather than `.Equals()` -- functionally identical for strings, and arguably cleaner in a LINQ `.Where()` chain. No action needed, just noting it's intentional and fine.

### Production impact -- as described

The silent false-pass in `OwlOntologyValidationTests.cs` (`ValidateMultilingualAnnotations` / `ValidateAIFMappings` early-returning "no concepts -> skip -> PASS") is a real correctness gap. This fix makes OWL validation reliable, which matters for the v0.9.0 validation gate.

PR #482, opened 2026-06-15
test(ontology): #204 OwlValidator live-path — silent false-pass is dead (dispatch 4rkh1s secondaire)

jsboige:
## OwlValidator live-path: silent false-pass is dead (dispatch `4rkh1s` secondaire)

**NEW additive file** — `OwlValidatorLivePathTests.cs` (+5 tests). Companion to PR #481 (the OwlAdapter reader fix). Dispatch `4rkh1s` **secondaire**: prove the production OWL validation path is LIVE.

### What it proves

The production validator `Tests/OwlOntologyValidationTests.cs` (a **runtime** validator invoked by `OwlValidatorConfig.Apply`, **NOT** an xUnit suite) had a **silent false-pass**:
- `ValidateMultilingualAnnotations` (L169-174) and `ValidateAIFMappings` (L220-225) both early-return `true` ("No concepts to validate — skipping") when `_ontology.GetResourcesByType(SKOSVocabulary.Concept)` returns empty.
- **Before #481**, that reader ALWAYS returned empty (RDFResource type-mismatch) → validators reported **PASS regardless** of whether any annotation or AIF mapping actually existed.

These tests drive the **REAL production validator** via reflection (its `_ontology` field is private; the validation methods are public). They prove the fix delivered the actual production delta:

| Test | Setup | Asserts |
|------|-------|---------|
| `Annotation_Validator_Inspects_Concepts_And_Passes_When_Annotated` | 2 fully-annotated concepts | PASS (genuine — annotations present, not skip-path) |
| `Annotation_Validator_Fails_When_Concepts_Lack_Annotations_After_Fix` | 4 concepts, zero annotations | **FAIL** (the check ran & found missing labels) — dead false-pass |
| `Aif_Validator_Fails_When_Concepts_Have_No_Match_Mappings_After_Fix` | concept with no match mappings | **FAIL** (check ran) |
| `Aif_Validator_Passes_When_Concepts_Carry_ExactMatch_After_Fix` | exactMatch mapping present | PASS (genuine) |
| `The_Reader_The_Validators_Branch_On_Resolves_Concepts_NonEmpty` | declared concepts | `GetResourcesByType(Concept).Count == 2` (the predicate that gated the false-pass) |

The **decisive** tests are the two FAIL assertions: before the fix they would have been PASS (empty concepts → skip). After #481 the concepts resolve, the check actually runs, and it correctly reports the missing annotations/mappings. That contrast IS the proof the false-pass is dead.

### Tests

223 pass / 0 fail / 5 skip (218 + 5). Deterministic, key-free, release-independent. No existing file modified. `AssetConverterConfig.cs` / `DatasetUpdaterRootConfig.cs` / `FallaciesLocalizationTests.cs` untouched (reserved #444 β/γ po-2023).

### Relation to #481

This PR depends on #481's reader fix (the validator tests assert non-empty concept resolution, which only holds after the fix). Recommend merge order: **#481 first, then this**. If #481 is reverted, tests #2/#3 here flip red (correctly — they'd expose the returning false-pass).

🤖 Worker po-2024 — dispatch `4rkh1s` secondaire.

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlAdapterRegressionTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlAdapterRegressionTests.cs
@@ -18,20 +18,22 @@ namespace Argumentum.AssetConverter.Tests.Ontology
     /// annotation scanners (<c>GetAnnotationSubjects</c> / <c>GetResourceAnnotations</c> /
     /// <c>GetLiteralAnnotations</c>).
     ///
-    /// ⚠️ These tests surfaced a REAL BUG (not greenwashed). Every fallback reader compares
-    /// <c>a.ValueIRI.Equals(value.URI)</c> / <c>a.SubjectIRI.Equals(subject.URI)</c> where
-    /// <c>.URI</c> is a <b>string</b> and <c>ValueIRI</c>/<c>SubjectIRI</c> are
+    /// ✅ BUG FOUND then FIXED. PR #480 surfaced a real bug (not greenwashed): every fallback
+    /// reader compared <c>a.ValueIRI.Equals(value.URI)</c> / <c>a.SubjectIRI.Equals(subject.URI)</c>
+    /// where <c>.URI</c> is a <b>string</b> and <c>ValueIRI</c>/<c>SubjectIRI</c> are
     /// <c>RDFResource</c>. <c>RDFResource.Equals(string)</c> returns <b>false by type-mismatch</b>,
-    /// so every read returns empty. See <see cref="Diag_RDFResource_Equals_String_Is_False_By_Type_Mismatch"/>.
-    /// This is the root cause behind the production validation module silently reporting
-    /// "no concepts → skip → PASS" on annotation/AIF checks. Reported as [BUG] on the dashboard.
+    /// so every read returned empty — the root cause behind the production validation module
+    /// silently reporting "no concepts → skip → PASS" on annotation/AIF checks. See
+    /// <see cref="Diag_RDFResource_Equals_String_Is_False_By_Type_Mismatch"/>.
     ///
-    /// Tests are split into: (A) characterization of the bug (pinned, documents current behavior),
-    /// (B) the correct comparison semantics (documents the fix), (C) write-path + serialization
-    /// (these DO work). No fix is applied — the file only pins observed behavior so a future fix
-    /// flips the [BUG] tests red→green and the fix author has a regression suite ready.
+    /// This PR applies the fix: drop <c>.URI</c> on the right-hand side of every reader comparison
+    /// (14 sites in <c>OwlAdapter.cs</c>), so readers compare <c>RDFResource.Equals(RDFResource)</c>.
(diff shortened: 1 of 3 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Solid companion PR to #481. The fix is mechanical and correct; the test coverage is thorough. A few observations:

---

### Production code (OwlAdapter.cs) - fix is correct

The core change across all 14 sites is identical in pattern: replace `.Equals(string)` with `.ToString() == .ToString()`. This works because `RDFResource.Equals(object)` dispatches by type and returns `false` when compared against a `string` or `Uri` - switching to string-string equality bypasses the type-mismatch.

Specific improvements beyond the raw `.ToString()` swap:
- **CheckIsNarrowerConcept**: The original `try { return ... } catch { return fallback }` pattern meant the fallback never executed if `SKOSHelper` returned `false` (no exception). The new `try { if (...) return true; } catch { } return fallback` correctly falls through on silent `false`, which is a real behavioral improvement - the annotation scanner now actually gets a chance to run.
- **GetExactMatchConcepts / GetCloseMatchConcepts / GetRelatedMatchConcepts**: Same fix - `SKOSHelper` returning `null` or empty no longer silently short-circuits; the annotation fallback now runs. This is the right pattern.

No issues with the production changes.

---

### Regression tests (OwlAdapterRegressionTests.cs) - well-flipped

All 8 section-(A) tests that previously asserted the broken behavior (`BeEmpty`, `BeFalse`) are now proper round-trip assertions. The assertions are specific (`HaveCount(3)`, `BeEquivalentTo(...)`, `ContainSingle(...)`) rather than just `NotBeEmpty`, which is good.

The `CheckIsNarrowerConcept` test now also asserts the negative case (`Unrelated` returns `false`), and `CheckHasClass` asserts `UndeclaredClass` returns `false`. Good defensive coverage.

---

### Live-path tests (OwlValidatorLivePathTests.cs) - well-designed with one structural concern

The test design is sound: prove the production validator actually *runs* (not just skips) by driving it through reflection. The 5 tests cover the two axes that matter:
1. **Annotated concepts -> PASS** (genuine, not skip-based)
2. **Unannotated concepts -> FAIL** (the silent false-pass is dead)
3. **AIF mappings -> same pattern for both PASS and FAIL**
4. **Contract guard** pinning `GetResourcesByType(Concept)` returns non-empty

**One concern - reflection fragility**: `BuildValidator` accesses a private `_ontology` field by name. If the field is renamed, these tests silently break (compile error, not a test failure - but still invisible in CI if the test project isn't rebuilt). This is acceptable for now since `OwlOntologyValidationTests` has no public injection seam, but worth noting that a longer-term fix should add a constructor overload or factory method that accepts an `OwlAdapter` directly. Not blocking.

**One minor observation**: `InvokeValidate` casts the return to `Task<bool>` without checking for `null`. Low risk but could be hardened.

---

### Security

No secrets, credentials, connection strings, or API keys in the diff.

### Verdict

Clean fix, thorough regression suite, well-documented. The reflection-based approach for the live-path tests is the right pragmatic choice. Ship it.

PR #485, opened 2026-06-15
test(parsing): #204 Fallacy ClassMap regression — Optional localized columns guard (+2)

jsboige:
## What

New regression coverage for `FallacyClassMap` via the real `CsvBase<Fallacy, FallacyClassMap>.LoadFromContent` path — **2 tests, additive only** (1 new file, +102).

Fallacy is the **largest entity (~1408 rows)** yet had **no** `LoadFromContent` regression coverage. The note in `EntityClassMapRegressionTests.cs` (#476) — _“Fallacy already has coverage elsewhere”_ — is **stale**: the only prior reference was an incidental `RegisterClassMap<FallacyClassMap>()` in `MmGeneratorTests` (which asserts no column mappings). Verified: `Fallacy.LoadFromContent` appears in zero tests before this PR.

## Tests

- **`Fallacy_LoadFromContent_MapsRequiredColumns`** — pins `pk` → both `PK` and `Id` (so `GetId()` resolves), the FR required columns (`Famille`, `TextFr`, `DescFr`, `ExampleFr`, `LinkFr`…), and that present localized columns (`text_en`, `desc_en`, `Family_zh`) map.

- **`Fallacy_OptionalLocalizedColumns_AbsentDoesNotThrow`** *(highest value)* — a **FR-only header set** (18 required columns, no `_en/_ru/_pt/_es/_ar/_fa/_zh/Latin/print_and_play/…`) parses **without throwing**, and every representative Optional column stays `null` across **all 8 languages + meta**. This guards the multilingual harvest: a single localized column losing its `.Optional()` would make `HeaderValidated` throw when loading a FR-only header — the exact fragility class of the #216 / #477 localization-mapping regression.

## Verification

- `dotnet test --filter FallacyClassMapRegressionTests` → **2/2 pass** (build green, only pre-existing warnings).
- Full suite runs in CI (Debug + Release).

## Scope

- **Additive only**: new file, no existing test or production code modified.
- Non-colliding with po-2024’s assigned `#204 CsvBase/MindMap` zone (this is the Entities/parsing zone).

🤖 Worker po-2023 — #204 test expansion, self-dispatched (standing directive « essayez d’avancer »; dispatch consumed, #457 deferred to ai-01).

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/Parsing/FallacyClassMapRegressionTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/Parsing/FallacyClassMapRegressionTests.cs
@@ -1,0 +1,102 @@
+using System.Linq;
+using Argumentum.AssetConverter.Entities;
+using FluentAssertions;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.Parsing
+{
+    /// <summary>
+    /// Regression tests for <see cref="FallacyClassMap"/> via the real load path
+    /// <see cref="CsvBase{T,TMap}.LoadFromContent"/>. Fallacy is the largest entity
+    /// (~1408 rows) and its ClassMap declares ~45 localized columns across 8 languages
+    /// (en/ru/pt/es/ar/fa/zh) plus Latin / print_and_play / AIF_* metadata — all .Optional().
+    ///
+    /// The critical guard: a single localized column losing its .Optional() would make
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** LGTM — 2 regression tests pinning FallacyClassMap optional column contract. Good doc explaining the fragility class (same as #216/#477). Inline CSV keeps tests self-contained. Required column mapping + optional absent guard both correct. Additive, no prod code touched.

PR #486, opened 2026-06-15
test(ontology): #133 e2e on real generated ontology — surfaces 2nd silent-false-pass bug (rdf:type dropped by OWL2XML round-trip)

jsboige:
## What

Completes the **#133 e2e OWL validation proof** by running the production validation path against the REAL generated ontology (`docs/ontology/argumentum.owl`). This is the real-data counterpart to #482's synthetic in-memory proof — and it **surfaces a second silent-false-pass bug** that the #480→#481→#482 lane missed (that lane only exercised the in-memory path).

## The finding — #133's premise does not hold for the prod load path

**The production OWL validator STILL silent-false-passes on any ontology loaded from a file.** The #481 fix killed the silent false-pass in-memory (proven by #482), but the real production path (`OwlOntologyValidationTests.LoadOntology` → `OwlAdapter.FromFile` → validate) is still broken, for a *different* root cause than the #480 `RDFResource.Equals` type-mismatch.

### Root cause (measured on the reloaded file)

OWLSharp's OWL2XML serializer **drops** the `rdf:type` and `skos:inScheme` annotation assertions during serialization — neither survives the round-trip:

| annotation | after reload |
|---|---|
| `rdf:type` | **0** (generator emitted ~1408) |
| `skos:inScheme` | **0** (generator emitted ~1408) |
| `skos:prefLabel` | 2816 ✅ survives |
| `skos:exactMatch/closeMatch/relatedMatch` | 10 ✅ survives |
| `skos:hasTopConcept` | 1 ✅ survives |

The OwlAdapter readers locate concepts/schemes by scanning `AnnotationAxioms` for `rdf:type`. On a **loaded** file `rdf:type` is absent → `GetResourcesByType(Concept)` returns empty → `ValidateMultilingualAnnotations` / `ValidateAIFMappings` hit their `if (concepts.Count == 0) return true;` guard → **report PASS without inspecting anything**. The ontology demonstrably contains 2816 prefLabels + 10 AIF matches — the validator just cannot see them.

## Tests (4, all green — characterization of current broken behavior)

1. `LoadedOntology_RdfTypeAndInScheme_DroppedByOwl2XmlRoundTrip` — pins `rdf:type==0`, `inScheme==0`, with `prefLabel>0` as contrast.
2. `LoadedOntology_ConceptAndSchemeReaders_ReturnEmpty_BugPinned` — `GetResourcesByType(Concept/ConceptScheme)` empty on loaded file.
3. `LoadedOntology_ContainsRealContent_ReadersCannotSee` — smoking gun: 2816 prefLabels + 10 matches + 1510 class declarations present; reader defect, not data defect.
4. `ProdValidator_SilentFalsePass_StillActiveOnLoadedOntology` — production validator returns TRUE for annotation + AIF checks on the loaded file (the silent false-pass, decisively pinned).

When the fix lands, assertions flip to the honest "detection works" form.

## Recommended fix (coordinator scope)

The readers must locate concepts via annotations that **survive** the round-trip — not `rdf:type`/`inScheme`. Options:
- subjects of `skos:prefLabel`/`skos:definition`/`skos:example` (literal-valued, survive), and/or
- filtered `DeclarationAxioms` (`OWLClass` minus the AIF/restriction/scheme classes), and/or
- locate the scheme as the subject of `skos:hasTopConcept`.

This is a **prod behavior change with release-gate implications** (validation that currently false-passes would start running for real), so it's left to coordinator scoping rather than a unilateral worker change. Filing as the #133 deliverable + bug report.

## Validation
- `dotnet test` full suite: **252 pass / 0 fail / 5 skip** (baseline 246 + #485 +2 + these 4).
- Deterministic, key-free, release-independent. NEW additive file — nothing modified.

cc @jsboige — this inverts #133's "confidence restored"; flagging before any release/validation decision.

🤖 Worker po-2024 (dispatch #133 primaire)

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlE2EGenerationValidationTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlE2EGenerationValidationTests.cs
@@ -1,0 +1,211 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using Argumentum.AssetConverter.Ontology;
+using FluentAssertions;
+using OWLSharp.Ontology;
+using RDFSharp.Model;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.Ontology
+{
+    /// <summary>
+    /// End-to-end OWL proof for #133 — SURFACES A SECOND SILENT-FALSE-PASS BUG in the production
+    /// validation path that the #480→#481→#482 lane missed.
+    ///
+    /// <see cref="OwlValidatorLivePathTests"/> (#482) proved with SYNTHETIC in-memory adapters that the
+    /// production validator no longer silent-false-passes after the OwlAdapter reader fix (#481). But
+    /// that lane only exercised the IN-MEMORY path. This class loads the REAL generation output
+    /// (<c>docs/ontology/argumentum.owl</c>, ~5 MB, produced by <c>OwlDocumentConfig.CreateOwlDocument</c>)
+    /// and proves the readers are STILL BLIND on a reloaded file — so the production validator
+    /// (<c>OwlOntologyValidationTests.LoadOntology</c> → <c>OwlAdapter.FromFile</c> → validate) STILL
+    /// silent-false-passes in the real prod path, for a DIFFERENT root cause than the #480 type-mismatch.
+    ///
+    /// ROOT CAUSE (measured on the loaded file): OWLSharp's OWL2XML serializer DROPS the
+    /// <c>rdf:type</c> and <c>skos:inScheme</c> annotation assertions during serialization — neither
+    /// survives the round-trip (rdf:type == 0, inScheme == 0 after reload). The OwlAdapter readers find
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Solid characterization test PR that surfaces a real bug: the OWL2XML round-trip silently drops `rdf:type` and `skos:inScheme` annotations from the generated ontology.

Key observations:
- Loads the real `docs/ontology/argumentum.owl` (~5MB), not a toy fixture — good fidelity
- Proves the drop mechanism end-to-end: write → OWL2XML serialize → read back → annotations gone
- Readers return empty collections where they should not, validators pass silently — this is the dangerous part
- Tests are correctly GREEN, pinning the *broken* behavior as a regression guard before the fix lands

One concern: `EntityReader.ParseEntityList()` has a documented `_skippable` field check that silently filters unknown annotations. The test confirms this masking behavior but does not instrument `_skippable` directly — worth a follow-up to confirm the fix will assert on skip-count, not just check final counts.

LGTM — this is the right way to surface a silent-false-pass before fixing it.

PR #489, opened 2026-06-16
fix(ontology): #133 OwlAdapter read-path survivor fallback — 2nd silent-false-pass dead

jsboige:
## What

Fixes the **second silent-false-pass** in the production OWL validation path — the one #486 characterized and the #480→#481→#482 lane missed because it only exercised the **in-memory** path.

**Root cause (verified by probing the reloaded ontology):** OWLSharp's OWL2XML round-trip drops `rdf:type` and `skos:inScheme` from the reloaded `AnnotationAxioms`. On a LOADED ontology, `GetResourcesByType(Concept)` / `GetConcepts()` scanned for `rdf:type` → **empty** → `ValidateMultilingualAnnotations` / `ValidateAIFMappings` hit their `if (concepts.Count == 0) return true;` guard → **PASS without inspecting**, even though the real content was present.

**Reloaded `AnnotationAxioms` breakdown (facts, not assumptions):**

| predicate | count | predicate | count |
|---|---|---|---|
| skos:prefLabel | 2816 | skos:broadMatch | 57 |
| skos:definition | 2816 | skos:closeMatch | 10 |
| skos:example | 2816 | skos:narrowMatch | 3 |
| skos:narrower / broader | 1407 | skos:hasTopConcept | 1 |
| **rdf:type** | **0** | **skos:inScheme** | **0** |

## Fix — READ-PATH ONLY (serializer untouched, per dispatch scope)

When the `rdf:type` scan is empty, locate entities via the SKOS annotations that DO survive the round-trip:
- **`skos:Concept`** → distinct subjects of `skos:prefLabel` (~1305 concepts resolved)
- **`skos:ConceptScheme`** → subject of `skos:hasTopConcept` (the scheme)

In-memory path is **preserved**: `rdf:type` is present in memory → early-return before the fallback → the #482 in-memory live-path proofs still hold unchanged. Concept subjects are deduped by **URI string** (not `RDFResource.Equals`) to sidestep the equality bug class of #480.

## Regression suite — #486 flipped red → genuine-green

| # | Test | Before (bug) | After (fix) |
|---|------|--------------|-------------|
| 1 | `…RdfTypeAndInScheme_DroppedByOwl2XmlRoundTrip` | drop real | drop real (now **benign**), prefLabel survives |
| 2 | `…ConceptAndSchemeReaders_ResolveViaSurvivingAnnotations…` | `BeEmpty()` | `NotBeEmpty()` (concepts >1000, scheme resolved) |
| 3 | `…ContainsRealContent_NowResolvableByReaders` | readers blind | readers resolve >1000 concepts |
| 4 | `…ProdValidator_InspectsConceptsAndGenuinelyPasses…` | false-pass (skip→true) | genuine pass (skip guard **unreachable**) |

## Verification
- Ontology namespace: **25/25 green** (e2e flipped + #482 in-memory live-path + #480 regression all hold)
- Full suite: **259 passed / 0 failed / 5 skipped** (no regression)
- Read-path only: **no write/serialize, no CSV/template/config touched**

## Dispatch
`msg-…irkf5i` primaire (base `4ac52e24`). Verdict/merge = ai-01 — I signal, I don't declare PASS.

🤖 Worker po-2024

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlE2EGenerationValidationTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlE2EGenerationValidationTests.cs
@@ -12,34 +12,31 @@ using Xunit;
 namespace Argumentum.AssetConverter.Tests.Ontology
 {
     /// <summary>
-    /// End-to-end OWL proof for #133 — SURFACES A SECOND SILENT-FALSE-PASS BUG in the production
-    /// validation path that the #480→#481→#482 lane missed.
+    /// End-to-end OWL proof for #133 — REGRESSION SUITE for the read-path fix of the SECOND
+    /// silent-false-pass bug (the one the #480→#481→#482 lane missed because it only exercised the
+    /// IN-MEMORY path).
     ///
-    /// <see cref="OwlValidatorLivePathTests"/> (#482) proved with SYNTHETIC in-memory adapters that the
-    /// production validator no longer silent-false-passes after the OwlAdapter reader fix (#481). But
-    /// that lane only exercised the IN-MEMORY path. This class loads the REAL generation output
-    /// (<c>docs/ontology/argumentum.owl</c>, ~5 MB, produced by <c>OwlDocumentConfig.CreateOwlDocument</c>)
-    /// and proves the readers are STILL BLIND on a reloaded file — so the production validator
-    /// (<c>OwlOntologyValidationTests.LoadOntology</c> → <c>OwlAdapter.FromFile</c> → validate) STILL
-    /// silent-false-passes in the real prod path, for a DIFFERENT root cause than the #480 type-mismatch.
+    /// HISTORY (the bug this file originally pinned, now fixed): OWLSharp's OWL2XML serializer/reader
+    /// drops <c>rdf:type</c> from the reloaded annotation stream (rdf:type==0, skos:inScheme==0 among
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Clean fix for the silent-false-pass bug characterized in #486. The approach is well-chosen: read-path only, serializer deliberately untouched (out of scope, OWLSharp concern).

**OwlAdapter.cs** (prod fix):
- `GetResourcesByType` gains a survivor fallback: rdf:type scan → empty → locate via surviving SKOS annotations (prefLabel for Concept, hasTopConcept for ConceptScheme). In-memory ontologies (rdf:type present) take the early-return and are unaffected — the #482 in-memory proofs still hold.
- `GetConcepts` similarly falls back to prefLabel subjects.
- New helper `GetAnnotationSubjectsByProperty` deduplicates by URI string (not RDFResource.Equals) — correct, since each concept carries fr+en = two assertions but one subject.
- The fallback is gated: only activates when the type scan returns 0, so normal in-memory paths are untouched. No perf concern.

**OwlE2EGenerationValidationTests.cs** (regression suite):
- Updated from characterization (pinning broken behavior → GREEN) to regression (asserting the fix → GREEN).
- 4 tests: (1) serializer drop still real, (2) readers now resolve on loaded file, (3) real content resolvable, (4) prod validator genuinely inspects (not skip-false-pass).
- Test 4 now asserts the concepts list is non-empty BEFORE calling the validator — proving the skip guard is unreachable. This is the right precondition check.

No security concerns. LGTM.

PR #495, opened 2026-06-16
test(ontology): #204 OwlAdapter survivor fallback contract — durcit #489/#133

jsboige:
## What

Unit-isolation regression suite for the **OWL2XML round-trip survivor fallback** in `OwlAdapter` — dispatch `qbw8vq` primaire (#204, **durcit #133 / #489**). One new additive test file, **0 production change**.

## Why

The #489 read-path fix (e2e-pinned on the real 5 MB ontology by `OwlE2EGenerationValidationTests`) made the production validators stop **silent-false-passing**: OWLSharp's OWL2XML serializer drops `rdf:type` from the reloaded annotation stream, so `GetResourcesByType` / `GetConcepts` — which locate entities by scanning for `rdf:type` — returned empty on any LOADED file, and the validators' `if (concepts.Count == 0) return true;` guard reported PASS without inspecting anything. The fix adds a **survivor fallback**: when the type scan is empty, resolve `skos:Concept` as the distinct subjects of `skos:prefLabel` and `skos:ConceptScheme` as the subject of `skos:hasTopConcept` (both survive the round-trip).

**The gap:** the survivor fallback branch (`OwlAdapter.cs` ~lines 424-441) was **unit-uncovered**. `OwlAdapterRegressionTests` (#480/#481) builds ontologies WITH `rdf:type` (via `DeclareConcept`), so `GetResourcesByType` always early-returns on the type scan and never reaches the fallback. The e2e covers the fallback on the real file but **coarsely** (`>1000`) and cannot pin the precise contracts below.

## What the e2e cannot pin (and this file does)

| Contract | Why the e2e can't | This file |
|---|---|---|
| **Exact dedup** | Real-file TextEn URI collisions aggregate subjects unpredictably → only `>1000` assertable | 3 concepts × fr+en prefLabel (6 assertions) → **exactly 3** distinct subjects |
| **Type-scan governs when present** | Real file has **zero** `rdf:type` → type scan is never the governing path | Mixed ontology (1 concept WITH `rdf:type`, 1 with ONLY `prefLabel`) → returns the type'd one **alone**, proving the fix didn't change in-memory semantics |
| **Unknown-type isolation** | e2e only queries Concept/ConceptScheme | Unknown type with survivor annotations present → **empty** (no over-resolution) |
| **ConceptScheme subject exactness** | e2e asserts only `NotBeEmpty` (real `hasTopConcept` count == 1) | Scheme resolves as the **subject** of `hasTopConcept` (not the object) |
| **AIF match survivor** | e2e inspects the whole match set | `exactMatch` + `closeMatch` annotation-scan fallback in isolation |

## Technique

A **survivor-only ontology** is built in-memory **without** `DeclareConcept` / `DeclareConceptScheme` (which emit `rdf:type`) — only the literal/resource annotations that survive the round-trip are added (`AnnotateConceptPreferredLabel`, `AnnotateConceptWithResource`). This reproduces the post-reload annotation graph **without loading a 5 MB file**, so the tests are fast (71 ms for the 7) and deterministic.

## Tests (7)

1. `Concept_ResolvedViaPrefLabelSubjects_WhenRdfTypeAbsent_Deduped` — 6 assertions → 3 distinct (the `.Distinct()` contract; a dropped dedup returns 6, a disabled fallback returns 0)
2. `ConceptScheme_ResolvedViaHasTopConceptSubject_WhenRdfTypeAbsent` — exact subject pin
3. `UnknownType_ReturnsEmpty_EvenWhenSurvivorAnnotationsPresent` — no over-resolution
4. `RdfTypePresent_TakesTypeScanPath_FallbackSkipped` — **decisive**: fix didn't change in-memory semantics
5. `GetConcepts_ResolvedViaPrefLabelSurvivorTail_WhenRdfTypeAbsent` — distinct reader path, same dedup
6. `ExactMatch_ResolvedViaAnnotationScan_WhenSkosHelperEmpty` — AIF survivor
7. `CloseMatch_ResolvedViaAnnotationScan_WhenSkosHelperEmpty` — pattern across match family

## Verification

```
Full suite: 282 passed / 0 failed / 5 skipped (287 total) — base was 275/0/5, +7
Ontology namespace: 32/32 pass (e2e 5MB + regression #480 + live-path #482 + this)
```

## Scope

- **Additive only**: 1 new file, no production code, no existing test modified.
- **0 gate risk**: pure test coverage.
- Builds on master `ac12e771` (#493 merged base).

Closes nothing on its own — **durcit** the #489 read-path fix so a regression of the survivor fallback surfaces at unit level (fast, deterministic) rather than only via the 5 MB e2e.

🤖 Worker po-2024 — dispatched by ai-01 (`qbw8vq`). Verdict QA visuelle / merge = ai-01.

Co-authored-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlAdapterSurvivorFallbackTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlAdapterSurvivorFallbackTests.cs
@@ -1,0 +1,252 @@
+using System.Linq;
+using Argumentum.AssetConverter.Ontology;
+using FluentAssertions;
+using OWLSharp.Ontology;
+using RDFSharp.Model;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.Ontology
+{
+    /// <summary>
+    /// Unit-isolation regression suite for the OWL2XML round-trip SURVIVOR fallback in
+    /// <see cref="OwlAdapter"/> — dispatch #204 primaire (durcit #133 / #489), cont. po-2024.
+    ///
+    /// The #489 read-path fix (pinned end-to-end on the real 5 MB ontology by
+    /// <see cref="OwlE2EGenerationValidationTests"/>) made the production validators stop
+    /// silent-false-passing: OWLSharp's OWL2XML serializer drops <c>rdf:type</c> from the reloaded
+    /// annotation stream, so <see cref="OwlAdapter.GetResourcesByType"/> /
+    /// <see cref="OwlAdapter.GetConcepts"/> — which locate entities by scanning for <c>rdf:type</c> —
+    /// returned empty on any LOADED file. The fix adds a SURVIVOR fallback: when the type scan is
+    /// empty, resolve <c>skos:Concept</c> as the distinct subjects of <c>skos:prefLabel</c> and
+    /// <c>skos:ConceptScheme</c> as the subject of <c>skos:hasTopConcept</c> (both survive the
+    /// round-trip). <see cref="OwlAdapter.GetExactMatchConcepts"/> and its match-family siblings carry
+    /// the same survivor shape (SKOSHelper-then-annotation-scan) for the AIF mappings.
+    ///
+    /// WHAT THIS FILE ADDS that the e2e suite cannot pin:
+    /// <list type="bullet">
+    /// <item>The survivor fallback BRANCH (<c>OwlAdapter.cs</c> lines ~424-441) is never reached by
+    /// <see cref="OwlAdapterRegressionTests"/> — those build ontologies WITH <c>rdf:type</c> (via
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

LGTM — solid regression suite for the OwlAdapter survivor fallback.

The 7 tests cover exactly what the e2e suite cannot:
- Fallback branch reachability (type scan empty → prefLabel subjects)
- Dedup contract (6 assertions → 3 distinct subjects)
- ConceptScheme via hasTopConcept subject (not object — subtle inversion pin)
- Unknown-type isolation (no over-resolution leak)
- Type-scan governance (rdf:type present → fallback skipped)
- GetConcepts() survivor tail (distinct path, same contract)
- AIF match-family (exactMatch + closeMatch annotation-scan fallback)

Zero production changes, additive test-only file. Well-documented rationale in the class doc explaining what each test pins that the e2e cannot.

No concerns — merge when ready.

PR #496, opened 2026-06-16
test(harvest): #204 extract expectedImageCount to pure method + 15 tests

jsboige:
## What

Extract the **CardPen-mirroring image-count formula** out of the `HarvestManager` Playwright flow into a pure, unit-testable static method `ComputeExpectedImageCount`, + 15 additive tests. Dispatch `qbw8vq` secondaire (#204). **Minimal output-neutral extraction** (the dispatch's "si propre" bar — judged clean) + tests.

## Why

The number of images CardPen generates per card set depends on its client-side rowset grouping. CLAUDE.md documents the formula in *"Leçons Apprises — Calcul du nombre d'images attendues"*:

```
with rsstyle="bunch" and rscount >= N:  expectedImageCount = ceil(cardIds.Count / rscount)
otherwise:                              expectedImageCount = cardIds.Count
```

A wrong value here surfaces at harvest time as the documented **"Mismatch image count"** debugging symptom — the harvest expects the wrong number of images and either hangs or returns a partial set. But the formula was **inlined inside the Playwright flow** (`HarvestManager.GenerateImages`), gated by `rscount > 1 && rsstyle in {bunch,cycle,random}`, with **zero unit coverage** — it could only be exercised by running a full browser harvest.

## The extraction (output-neutral)

`HarvestManager.ComputeExpectedImageCount(int cardCount, int rscount, string rsstyle)` is a pure static method replicating the exact inline formula. The call site change is **output-neutral**:

```diff
-                  expectedImageCount = cardIds.Count;
+                  expectedImageCount = ComputeExpectedImageCount(cardIds.Count, rscount, rsstyle);

                   if (rscount > 1 && !string.IsNullOrEmpty(rsstyle) &&
                       (rsstyle == "bunch" || rsstyle == "cycle" || rsstyle == "random"))
                   {
-                      expectedImageCount = (int)Math.Ceiling((double)cardIds.Count / rscount);
                       Log($"[rscount adjustment] rscount={rscount}, ...");
                   }
```

- `expectedImageCount` gets the **same value** (the method replicates the exact `ceil`/`rscount` formula).
- The **Log condition is preserved verbatim** (`rscount > 1 && rsstyle valid`) — same log line, same trigger.
- Production build: **0 errors**.

## Tests (15) — `HarvestManagerExpectedImageCountTests`

**13-row Theory matrix** covering the full input space:
- Grouping applies (`rscount > 1` + valid lowercase `rsstyle`) → `ceil`: exact-divide (6/3=2), non-divide (ceil(7/3)=3), Memo (200/200=1), single-card (ceil(1/3)=1), `cycle`, `random`
- Grouping does NOT apply → `cardCount` verbatim: `rscount==1`, `rscount==0`, empty `rsstyle`, null `rsstyle`, unknown `rsstyle`, case-sensitivity (`"BUNCH"`, `"Bunch"`)

**Memo golden-master Fact** — the CLAUDE.md headline example: 200 CSV rows × `rscount=200` bunch → exactly 1 image. A regression to 200 or 0 breaks the Memo harvest.

**Case-sensitivity Fact** — CardPen `rsstyle` is lowercase-only; `"Bunch"`/`"BUNCH"`/`"bUnCh"` must NOT group.

## Verification

```
Full suite: 297 passed / 0 failed / 5 skipped (302 total) — was 282/0/5, +15
Production build: 0 errors (16 pre-existing warnings, all SGEN/nullable — unrelated)
```

## Scope

- **Output-neutral prod change** (1 method extracted + call site, +21/-2 lines in `HarvestManager.cs`) + 1 new test file.
- Builds on master `ac12e771`.
- Separate scope from #495 (ontology survivor fallback) — different subsystem, different branch.

Closes nothing on its own — **pins** a documented fragile contract that had 0 unit coverage, so a regression of the CardPen-mirroring formula surfaces in 43 ms rather than only via a full browser harvest.

🤖 Worker po-2024 — dispatched by ai-01 (`qbw8vq`, secondaire). Verdict QA visuelle / merge = ai-01.

Co-authored-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/HarvestManagerExpectedImageCountTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/HarvestManagerExpectedImageCountTests.cs
@@ -1,0 +1,98 @@
+using Argumentum.AssetConverter;
+using FluentAssertions;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.WebBasedGenerator
+{
+    /// <summary>
+    /// Contract pin for <see cref="HarvestManager.ComputeExpectedImageCount"/> — dispatch #204
+    /// secondaire (cont. po-2024).
+    ///
+    /// The number of images CardPen generates for a card set depends on its client-side rowset
+    /// grouping. CLAUDE.md documents the formula in "Leçons Apprises — Calcul du nombre d'images
+    /// attendues":
+    /// <code>
+    ///   with rsstyle="bunch" and rscount &gt;= N:  expectedImageCount = ceil(cardIds.Count / rscount)
+    ///   otherwise:                                expectedImageCount = cardIds.Count
+    /// </code>
+    /// A mismatch here surfaces at harvest time as a confusing "image count mismatch" (one of the
+    /// documented debugging symptoms in CLAUDE.md) — the harvest expects the wrong number of images
+    /// and either hangs or returns a partial set.
+    ///
+    /// This formula was previously INLINED inside the Playwright flow in
+    /// <see cref="HarvestManager"/> (GenerateImages), gated by <c>rscount &gt; 1 &amp;&amp; rsstyle in
+    /// {bunch,cycle,random}</c>, with ZERO unit coverage — it could only be exercised by running a
+    /// full browser harvest. It has been extracted (output-neutral — the call site preserves the
+    /// exact computation and the Log condition verbatim) into the pure, deterministic
+    /// <see cref="HarvestManager.ComputeExpectedImageCount"/> so the CardPen-mirroring contract is
+    /// unit-testable. These tests pin the contract additively.
+    /// </summary>
+    public class HarvestManagerExpectedImageCountTests
+    {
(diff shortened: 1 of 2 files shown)


clusterManager-Myia:
**[NanoClaw]**

test(harvest): extract `ComputeExpectedImageCount` to pure method + 15 tests.

**Extraction:** The inline `ceil(cardCount/rscount)` formula is pulled into a `public static` method. The call site in `GenerateImages` now calls `ComputeExpectedImageCount(cardIds.Count, rscount, rsstyle)` — same computation, same guards (`rscount > 1`, non-empty rsstyle in `{bunch, cycle, random}`), output-neutral. The `Log` call still fires inside the guard block (unchanged behavior).

**Tests:** 15 Theory cases covering: even/uneven division, CLAUDE.md Memo golden-master (200 rows/rscount=200 → 1 image), single-card edge, all three styles, and non-grouping guards (rscount <= 1, empty/null/unknown rsstyle, case-sensitivity). Plus 2 dedicated Fact tests (golden-master pin, case-sensitivity contract).

No secrets, no issues. LGTM.

PR #500, opened 2026-06-16
test(printplay): #204 extract recto-verso back reorder to pure method + 9 tests

jsboige:
## What & why

The Print & Play PDF (`PrintAndPlayDocument`) prints card backs on the reverse of each sheet so a sheet can be flipped to read the back. To make each back line up behind its matching front when the sheet is turned along its horizontal edge, the backs of **each grid row** are reversed before rendering — a per-row reversal (not a full-array mirror), because a horizontal flip swaps left/right *within* each row while preserving row order.

This fragile alignment contract was **inlined in `Compose()` with zero unit coverage**:

```csharp
var backCardsArray = pageBackImages.ToJaggedArray(nbColumns)
    .Select(row => row.Reverse().ToArray()).ToArray().Flatten();
```

It could only be exercised by rendering a full QuestPDF document and visually inspecting the back alignment. A regression here (e.g. reversing the *whole* array instead of per-row, or flipping rows too) produces printed sheets whose backs mismatch their fronts — **unusable sheets, only caught at print time**.

## Change

**Output-neutral extraction + additive test coverage.** No behavior change.

1. Extracted the inline composition into a pure, deterministic, generic static method:
   ```csharp
   public static T[] ReorderBacksForRectoVerso<T>(IList<T> backs, int nbColumns)
       => backs.ToJaggedArray(nbColumns)
               .Select(row => row.Reverse().ToArray())
               .ToArray()
               .Flatten();
   ```
   The call site (`Compose()`) now reads `ReorderBacksForRectoVerso(pageBackImages, nbColumns)` — identical computation, just unit-testable.

2. Added `PrintAndPlayRectoVersoContractTests` (9 tests) pinning the contract:
   - **Headline** — full rows: 6 backs / 3 cols → `[B2,B1,B0,B5,B4,B3]` (per-row reversal, row order preserved; explicitly rejects a full-array mirror `[B5..B0]`).
   - Trailing short row — the lone card on a partial row reverses to itself and stays at the tail.
   - Single column — horizontal flip is identity (1-wide row reverses to itself).
   - One short row / width-2 / empty / single-back — degenerate cases (no crash, correct identity).
   - **byte[] grounding** — proves the generic works for the production element type (card image bytes), not just string labels.
   - **Cell-wise alignment semantic** — rebuilds input + output grids and asserts `output[row][col] == input[row][cols-1-col]` for every cell (the precise within-row mirror).

The building blocks (`ToJaggedArray`, `Flatten`) are already pinned by `UtilityExtensionsLayoutTests`; this PR pins the **Print&Play-specific composition** — the gap that test file does not cover.

## Verification

- New class: `9/9` pass.
- Full suite on clean master base: **291 passed / 0 failed / 5 skipped** (5 skips = Freeplane GUI + others requiring interactive session) — **no regression**.
- Diff: +1 pure static method (output-neutral) + 1 test file. No CSV, no config, no `RowsetNb`/`rscount` changes, no workflow/rules touched.

## Scope

Dispatch `#204` γ — test-coverage expansion on fragile pipeline contracts. Pivot sanctioned by dispatch `qbw8vq` ("sinon pivote PdfManager layout math"). Additive only; no production behavior change.

🤖 Worker po-2024

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/PrintAndPlayRectoVersoContractTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/PrintAndPlayRectoVersoContractTests.cs
@@ -1,0 +1,206 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Argumentum.AssetConverter;
+using FluentAssertions;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.WebBasedGenerator
+{
+    /// <summary>
+    /// Contract pin for <see cref="PrintAndPlayDocument.ReorderBacksForRectoVerso{T}"/> — dispatch #204
+    /// tertiaire / γ (cont. po-2024).
+    ///
+    /// The Print &amp; Play PDF (<see cref="PrintAndPlayDocument"/>) prints card backs on the reverse of
+    /// each sheet so a sheet can be flipped to read the back. To make each back line up behind its
+    /// matching front when the sheet is turned along its horizontal edge, the backs of EACH GRID ROW
+    /// are reversed before rendering. This is a per-row reversal of the back grid — NOT a full mirror
+    /// of the flat array — because a horizontal flip swaps left/right WITHIN each row while preserving
+    /// row order (row 0 stays row 0). In other words the back at output position (row, col) is the
+    /// original back at (row, nbColumns-1-col).
+    ///
+    /// This reordering was previously INLINED inside <see cref="PrintAndPlayDocument.Compose"/> as the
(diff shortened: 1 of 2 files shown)


clusterManager-Myia:
**[NanoClaw]**

test(printplay): extract recto-verso back reorder to pure generic method + 8 tests.

**Extraction:** The inline `backs.ToJaggedArray(nbColumns).Select(row => row.Reverse().ToArray()).ToArray().Flatten()` is pulled into a generic `ReorderBacksForRectoVerso<T>` — output-neutral, call site preserves exact computation.

**Tests:** 8 Fact tests covering: full evenly-divisible rows, trailing short row, single column, short row within larger column count, even width-2, empty backs, single back, production byte[] element type, and a cell-wise alignment contract (output[row][col] == input[row][cols-1-col]). Well-documented doc comments explain the per-row reversal semantics vs full mirror.

No secrets, no issues. LGTM.

PR #472, opened 2026-06-14
perf(#29): 3 output-neutral cleanups (held — do not merge during gate)

jsboige:
## Summary

3 **LOW-risk, output-neutral** perf cleanups from the [#29 memory/perf audit](../blob/master/docs/investigations/2026-06-14-29-memory-perf-hotspots.md) §3 shortlist. Dispatched by ai-01 (`ou00g3` v2) as the primary key-free task.

## Changes
| # | File | Fix | Output impact |
|---|------|-----|---------------|
| **H6** | `UtilityExtensions.cs` | `new HttpClient()` per call → `static readonly _sharedHttpClient` | none (same payloads) |
| **H5** | `PrintAndPlayDocument.cs` | `File.ReadAllBytes(header)` per page → `Lazy<byte[]>` cached once | none (same bytes) |
| **H7** | `HarvestManager.cs` | remove dead `consoleMessages` param (never written — confirmed by grep in both call sites + GenerateImages body) | none |

## Proof of output-neutrality
- Build **green** (0 errors; 18 warnings all in untouched files).
- Tests **159 passed / 0 failed / 5 skipped** — matches baseline exactly (zero regression).
- No render/CSV/PDF logic touched; no MED/HIGH hotspots (H1-H4/H8 need profiling/design — out of scope).

## ⚠️ HELD — do NOT merge during gate
Per dispatch rule: this PR is open but **not self-merged**. ai-01/jsboige decide timing post-validation (master untouched while jsboige validates the 64 Release PDFs on po-2023). The branch does not affect validation (jsboige validates already-regenerated artefacts).

🤖 Worker po-2024 (dispatch `ou00g3` primary)

PR #504, opened 2026-06-16
test(localization): #204 #216 field-mapping contract — pure extraction + 12 tests

jsboige:
## What

Primaire of ai-01 dispatch `bwpx5q` (gate-safe #204 deep-queue). Extracts the **#216 root-cause site** — the field-conversion transform inside `CardSetLocalization.TranslateCardSetInfo` — into pure, output-neutral static methods, and pins its contract with 12 tests.

**#216 in one line:** `LocalizationConfig.FrontFieldConversions` did `template.Replace(sourceField, destField)` on field names that were *absent* from the Mustache template. `string.Replace` is silent when the pattern is missing (returns the string unchanged), so every non-FR PDF silently kept its French content with no error — a silent false-pass. This PR makes that contract directly unit-testable and adds the regression guard that would have caught the original bug.

## How (output-neutral, gate-safe)

Two new `public static` methods on `CardSetLocalization`:

- **`ApplyFieldConversions(template, fieldConversions, exceptionPatterns, destLanguage) → string`** — byte-identical reimplementation of the previous inline logic (field-token replace + exception backtrack). `TranslateCardSetInfo` now calls it (3 lines replacing 31). `FormatField` delegates to a shared `FormatFieldToken` so production + the pure helpers + the 4 existing localization test files all share one source of truth.
- **`FindAbsentSourceFields(template, fieldConversions, destLanguage) → IReadOnlyList<string>`** — the **fail-loud companion**: returns the source field names whose token is *not* in the template (the ones `ApplyFieldConversions` would silently skip). Pure capability.

### Why `FindAbsentSourceFields` is not yet wired to `Logger`

The dispatch's DoD asked for "champ absent → fail-loud/log, PAS silencieux", but the gate holds *even output-neutral pipeline behaviour changes* (cf #472). Wiring `Logger.Log` on absent fields would emit new log lines in the production path = a behaviour change. So this PR delivers the **capability + contract** (the pure method + tests) and defers the **1-line production wiring** to a post-gate follow-up. The produced template is byte-identical to today → unambiguously output-neutral.

If ai-01 wants the logging live *now* (lifting the gate for it), it's a one-line addition in `TranslateCardSetInfo`.

## Tests — `FieldMappingContractTests` (12)

- **(a) present → replaced:** `ApplyFieldConversions_PresentField_IsReplaced`, `..._ReplacesAllOccurrences_NotJustFirst` (global, not first-match).
- **(b) absent → hazard + fail-loud:** `..._AbsentField_IsSilentlySkipped_NoThrow` (pins the silent-skip as the #216 hazard), `FindAbsentSourceFields_ReportsAbsentField_AndOmitsPresentOne`, `..._ReturnsEmpty_WhenEveryFieldIsPresent`.
- **dest-lang edges:** `ApplyFieldConversions_NoConversionForDestLang_SkipsSilently`, `FindAbsentSourceFields_NoConversionForDestLang_IsNotReportedAbsent` (a field with no `en` entry is out of scope, not "absent").
- **(c) 4-CardSet lockstep (the #216 regression guard):** `EveryMappedFrontField_IsPresentInRealTemplate_NoSilentFalsePass` — Theory over Fallacies / Virtues / Rules / Scenarii, each against its real representative template (`Argumentum_Fallacies_Face_fr.json`, `Argumentum_Virtues_Face_fr.json`, `Argumentum_Rules_fr.json`, `Argumentum_Scenarii_Face_fr.json`). If any CardSet ever gains a front-field mapping whose source name isn't in the template, this fails loud — the signal #216 lacked.
- **grounding:** `ApplyFieldConversions_OnRealFallaciesTemplate_ReplacesFrenchPlaceholders` — the pure method fed the real config + template performs the same `text_fr → text_en` fix as the production path (behaviour-preserving, not just structurally pure).

## Verification

- New class: **12/12 pass**.
- Full suite on branch: **318 passed / 0 failed / 5 skipped** (306 post-#500 master baseline + 12 new). **No regression** — the 4 existing localization test files (`Fallacies/Rules/Virtues/Scenarii LocalizationTests`) still pass through `TranslateCardSetInfo` → `ApplyFieldConversions`, confirming byte-identical output.
- Diff: `CardSetLocalization.cs` (extract + delegate, net +~85/-25), 1 new test file. No CSV, no config, no `RowsetNb`/`rscount`, no workflow/rules, no production `Logger` change.

## Scope notes

- The 4 sibling test files each re-implement the apply loop inline (`loc.FormatField` + manual `Replace`). Migrating them onto `ApplyFieldConversions` (single source of truth) is a clean follow-up — out of scope here to keep this PR a focused contract extraction.
- This is **test-only / output-neutral** → mergeable during the release gate per ai-01's gate-safe framing.

Contributes to #204. Refs #216.

🤖 Worker po-2024 — primaire of dispatch `bwpx5q`.

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/Localization/FieldMappingContractTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/Localization/FieldMappingContractTests.cs
@@ -1,0 +1,203 @@
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using FluentAssertions;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.Localization
+{
+	/// <summary>
+	/// Contract tests for the #216 root cause, pinned via the pure extraction
+	/// <see cref="CardSetLocalization.ApplyFieldConversions"/> /
+	/// <see cref="CardSetLocalization.FindAbsentSourceFields"/>.
+	///
+	/// #216 in one line: <c>LocalizationConfig.FrontFieldConversions</c> did
+	/// <c>template.Replace(sourceField, destField)</c> on field names that were ABSENT from the
+	/// Mustache template. <c>string.Replace</c> is silent when the pattern is missing (returns the
+	/// string unchanged), so every non-FR PDF silently kept its French content with no error — a
+	/// silent false-pass. These tests pin the two halves of the contract:
+	///   (a) a mapping whose source field EXISTS in the template → it is replaced;
+	///   (b) a mapping whose source field is ABSENT → it is silently skipped by ApplyFieldConversions
+	///       (the hazard) BUT surfaced by FindAbsentSourceFields (the fail-loud companion);
+	///   (c) Fallacies/Rules/Virtues/Scenarii lockstep — every mapped source field must be present
+	///       in a real template (the regression guard that would have caught #216).
+	///
+	/// The four sibling test files (Fallacies/Rules/Virtues/Scenarii LocalizationTests) each
+	/// re-implement the apply loop inline; this file drives the shared pure methods directly.
+	/// Migrating those helpers onto <see cref="CardSetLocalization.ApplyFieldConversions"/> is a
+	/// follow-up (single source of truth), out of scope for this gate-safe extraction.
+	/// </summary>
+	public class FieldMappingContractTests
+	{
+		private static readonly string RepoRoot = FindRepoRoot();
+
+		private static string FindRepoRoot()
+		{
+			var dir = new DirectoryInfo(AppContext.BaseDirectory);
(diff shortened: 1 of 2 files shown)


clusterManager-Myia:
**[NanoClaw]** LGTM — test + refactor, 2 files (+297/-32)

Clean extraction of the #216 root cause: `ApplyFieldConversions` and `FindAbsentSourceFields` pulled out as pure static methods from inline logic (output-neutral, byte-identical). 12 contract tests covering the silent false-pass hazard (field absent from template → `string.Replace` silently returns unchanged). All 4 card sets (Fallacies/Rules/Virtues/Scenarii) pinned. Good regression guard.

No concerns.

PR #490, opened 2026-06-16
feat(dnn-i18n): #457 translate DNN UI strings FR -> 7 langs via gpt-5.5

jsboige:
## What

Populates the 7 target-language columns (en/ru/pt/es/ar/fa/zh) for the DNN web-platform UI strings — **63 cells across 9 rows**, leaving source data untouched (`SelectEmptyTargets=true`, 0 overwrites). Closes the data deliverable of #457.

## How

DatasetUpdater run via **OpenRouter `openai/gpt-5.5`** (function calling). 

> **Provider note:** The merged #487 task config points at the OpenAI direct key (`openai-key.txt`, bare `gpt-5.5`, no `BaseUrl`) — but that key is currently **HTTP 429 `insufficient_quota` (exhausted)**. Per the #457 dispatch ("Clé OpenRouter"), I ran the translation through OpenRouter instead: `Model=openai/gpt-5.5`, `OpenAIKeyPath=.keys\openrouter-key.txt`, `BaseUrl=https://openrouter.ai/api/v1`. WebSearch confirmed `openai/gpt-5.5` exists on OpenRouter **with tool-calling support** (required for the `UpdateRecord` function). The smoke (2 rows, 14 calls) and the full run (9 rows, 63 calls) both **exit 0, no 429** → the sk-or- key has credits. **gpt-5.5 confirmed — NOT 5.4-mini.**

A follow-up config PR could switch the merged #487 task to the OpenRouter path so the run is reproducible on master without the exhausted OpenAI key — flagged for @jsboige's provider decision (recharge OpenAI vs. adopt OpenRouter).

## Rows

| Row | FR source | Translated? |
|-----|-----------|-------------|
| `ui.fallacy.find_out_more` | solid | ✅ 7 langs |
| `ui.rules.players_range` | solid | ✅ 7 langs (placeholders `{0}/{1}` preserved) |
| `res.RuleSummary` / `RuleMaterial` / `RuleInstallation` / `RuleVariants` / `RuleMemoCard` / `RuleMemoCardFileNamePrefix` / `RuleMemoCardDownload` | ⚠️ **INFERRED FR** | ✅ 7 langs — **verify FR vs 2sxc App Resources DB export before relying on the 7 translations** |
| `res.RuleMemoInstructions` | ❌ none | ⏭️ **SKIPPED** — DB-only multi-sentence text, export required before translation |

## Quality

- `{0}`/`{1}` placeholders preserved at equivalent position in every language
[...]

PR #487, opened 2026-06-15
feat(dataset-updater): #457 DNN UI strings localization infra (Enabled=false)

jsboige:
## What

DatasetUpdater **config infrastructure** for issue **#457** — translate the DNN web-platform UI strings (`docs/dnn-localization/dnn-ui-strings.csv`) FR → all 7 languages (en/ru/pt/es/ar/fa/zh) via **gpt-5.5**. The task ships **`Enabled = false`** so the mechanics can be reviewed before the smoke/full run.

This is the **config-infra PR** (my ACK to ai-01, `rqb5pm`). The smoke test + full run → **data PR** is the next step.

## Files (7, +320/-13)

| File | Change |
|------|--------|
| `Entities/DnnUiString.cs` | NEW — entity + `DnnUiStringClassMap` (`key`=PK, only required col) |
| `WebBasedGenerator/KnownDataSets.cs` | + `DnnUiStrings` constant |
| `AssetConverterConfig.cs` | + `DataSetInfo` (DebugFilePath → `docs/dnn-localization/dnn-ui-strings.csv`, CsvType=`DnnUiString`) |
| `DatasetUpdater/DatasetUpdaterRootConfig.cs` | + task block (Enabled=false, gpt-5.5, OpenRouter key, ChunkSize=4, multi-lang single task) |
| `Resources/PromptDnnUiStringsTranslateMultiUser.txt` | NEW prompt |
| `Resources/PromptDnnUiStringsTranslateMultiAssistant.txt` | NEW prompt |
| `Tests/Parsing/DnnUiStringClassMapRegressionTests.cs` | NEW +4 tests |

## Mechanics note for ai-01 (review)

- **DUAL CSV PATH.** The engine reads the source via `DataSetInfo.GetDictionaryFromCsv`, which uses a **plain** `CsvConfiguration` (no `PrepareHeaderForMatch`, **no ClassMap**) → dictionary keys are the **raw CSV header names**. Therefore `FieldsToInclude`/`FieldsToUpdate`/`PrimaryField` use the literal headers (`key`, `fr`, `en`, `source_file`…), **not** property names. The `DnnUiStringClassMap` is only on the harvest/validation path (`CsvBase.LoadFromContent`) and is exercised by the 4 new tests. Verified by reading `DataSetInfo.cs:396-410` + `DatasetUpdaterConfig.cs:125-157`.
[...]

PR #524, opened 2026-06-17
tools(dnn-i18n): #457 content-type extractor + reimport dry-run (fixture, Option C tooling)

jsboige:
## What

**#457 DNN i18n tooling** — bricks 1 + 3 (content-type extractor + re-import dry-run verifier) + round-trip DoD test. Dispatched by ai-01 (`msg-…141408`, non-gated #457 tooling while #131 exec waits on jsboige's target decision).

## Key finding that shaped the work

**Brick 2 (DatasetUpdater Option C config) was ALREADY merged in #487** (`ca9a8640`) — entity `DnnUiString`, `KnownDataSets.DnnUiStrings`, prompts, task config (`DatasetUpdaterRootConfig.cs:2636-2696`, `Enabled=false`, gpt-5.5). The investigation (SDDD triple-grounding) confirmed the config-only gpt-5.5 path is canonical. So this PR adds the **missing tooling *around* it** rather than rebuilding the config:

| Brick | What | Status |
|-------|------|--------|
| (2) config Option C | gpt-5.5 task ui.*/res.* | ✅ already done (#487) |
| **(1) extractor** | content-type → CSV (codifies PHASE1 manual audit) | 🆕 net-new |
| **(3) re-import verifier** | key-set diff + payload render (dry-run) | 🆕 net-new |

## Brick 1 — `extract_dnn_ui_strings.py`

Parses the Argumentum 2sxc `.cshtml` templates, emits the `dnn-ui-strings.csv` dialect (`key,context,source_file,fr,en,ru,pt,es,ar,fa,zh,notes`):
- **`ui.*`** = anchor-based (NOT free-text): each entry declares a verbatim anchor that must exist in the named source. Fails **loud (exit 2)** if a refactor removes a string — never silently drops a row, never fabricates.
- **`res.*`** = `@?Resources.<Key>` regex capturing both `@Resources.X` (directive) and `@Html.Raw(Resources.X)` (nested expression). The nested form recovers `RuleMemoInstructions` (the one a naïve `@Resources.` regex misses). `fr` left empty + flagged `DB-only` (canonical values live in SQL/2sxc App Resources — pattern appris).

## Brick 3 — `reimport_dnn_ui_strings.py`

- `verify` — key-set diff vs reference. **KEY-SET match is HARD** (exit 1 on divergence). `res.*` fr empty-in-extract is by-design WARNING.
- `reimport` — renders the canonical DNN/2sxc payload to **stdout only** (never writes — live apply is DB/RDP-gated, jsboige).

## DoD proof — `test_roundtrip.py`

```
[1/4] extract OK: 10 rows (2 ui.*, 8 res.*)
[2/4] key set OK (10 keys)
[3/4] verify OK (key sets match, res.* fr empty by design)
[4/4] reimport dry-run OK (payload rendered, nothing written)
ROUND-TRIP DoD PASS — exit 0, zero prod mutation
```

## Anti-fabrication (proven, not asserted)

- **Negative test**: breaking a `ui.*` anchor → exit 2 (verified, not piped).
- **Cross-validated vs REAL prod templates**: extractor run on `DNNPlatform/Portals/1/2sxc/Argumentum/` yields **10/10 keys** matching `docs/dnn-localization/dnn-ui-strings.csv`. Only delta = `res.*` fr (empty-in-extract by design — DB-only).

## Gate boundaries (HARD)

- ❌ No live DNN DB / portal / 2sxc App Resources mutation (DB/RDP-gated jsboige).
- ❌ Does NOT touch `dnn-ui-strings.csv` (po-2024 lane, #490).
- ❌ Does NOT enable/modify the DatasetUpdater config (#487 stays `Enabled=false`).
- ❌ No translation run (gpt-5.5 translation is the config's job, gated on FR-complete which needs the portal export).
- ❌ No QA verdict declared (ai-01 only). Stdlib Python, 0 deps.

### DIFF
--- a/tools/dnn_i18n/README.md
+++ b/tools/dnn_i18n/README.md
@@ -1,0 +1,79 @@
+# #457 DNN site localization — i18n tooling (bricks 1 + 3 + round-trip)
+
+Reusable tooling around the **already-wired** DNN UI-strings translation rail (DatasetUpdater
+Option C, config merged in #487, `Enabled=false`). These tools do the work *around* the config
+that the config can't do for itself: extract the content-type set from the templates, and
+dry-run-verify the re-import — all **fixture/dry-run**, zero prod mutation.
+
+## The 3 bricks (and which already existed)
+
+| Brick | What | Status |
+|-------|------|--------|
+| **(2) DatasetUpdater config (Option C)** | gpt-5.5 task for `ui.*`/`res.*`, 8-language | ✅ **Already done** — #487 (`ca9a8640`), `DatasetUpdaterRootConfig.cs:2636-2696`, `Enabled=false` |
+| **(1) content-type → CSV extractor** | reusable, codifies the PHASE1 manual audit | 🆕 net-new — `extract_dnn_ui_strings.py` |
+| **(3) CSV → DNN re-import dry-run verifier** | key-set diff + payload render (no write) | 🆕 net-new — `reimport_dnn_ui_strings.py` |
+
+**Brick 2 was NOT rebuilt** — the investigation confirmed the config-only gpt-5.5 path is already
+canonical (entity `DnnUiString`, `KnownDataSets.DnnUiStrings`, prompts, task config all merged,
+`Enabled=false`). Re-adding it would duplicate. This PR adds the missing tooling (bricks 1 + 3).
+
+## Files
+
+- `extract_dnn_ui_strings.py` — parse 2sxc `.cshtml`, emit `dnn-ui-strings.csv` dialect.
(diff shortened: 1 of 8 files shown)


clusterManager-Myia:
**[NanoClaw]**

Clean i18n tooling (bricks 1+3) for DNN localization: anchor-based ui.* extractor (anti-fab: fails loud if anchor disappears, never invents rows), res.* reference scanner (DB-only values honestly flagged), CSV reimport dry-run verifier, and a round-trip test with fixture. Stlib-only Python, cross-validated against prod templates (10/10 key match). Gate boundaries explicit: no DB touch, no live reimport, no CSV modification, no DatasetUpdater config change. Subprocess only used to run its own test script. Good README with anti-fabrication guarantees. LGTM.

PR #555, opened 2026-06-20
docs(dnn): #506 social-auth connectors inventory + update procedure

jsboige:
Repo-grounded inventory of the social-auth providers shipped with the DNN 9.11.1 export + the update/secret-rotation procedure.

## What this covers
- **§1 inventory** (repo-grounded, no live access): 5 provider source dirs (DNN/Facebook/Google/Live/Twitter), 4 stock assemblies (9216 bytes each = stock 9.11.1), JWT auth `DnnJwtAuth_09.11.01`.
- **§2 highest-risk finding**: the Microsoft provider ships as **`LiveConnect`** (`Live/`) — the legacy pre-Entra name, talking to deprecated `login.live.com` endpoints. Most likely to already be broken. Recommend NOT patching on 9.11.1; fold into #131 upgrade.
- **§3 per-provider deprecation table**: Graph API version, Google Identity Services migration, Entra v2.0, Twitter API tier — with the developer console + 'healthy' criteria each.
- **§4 coordination with #131**: sequence pre-upgrade inventory → during-upgrade sandbox retest → post-upgrade prod rotation.
- **§5 secret-rotation procedure** (5 steps) with the hard rule **NO secret in repo** — App ID/secret live only in DNN admin UI / server-side.
- **§6 open questions for jsboige** (which providers actually used, Microsoft keep/retire, Twitter keep/retire, timeline).
- **§7 out of scope** explicit.

## Why this now
ai-01 dispatched 'Phase 3 (2sxc 15→21) OR #506 connecteurs auth' for po-2023 during the RDP-blocked regen wait. The 2sxc migration is **already fully documented** (PR #543), so #506 is the non-duplicate maillon. Doc/research-only — no prod mutation, no secret stored.

Index README.md (#554) updated to reference the new doc.

Dispatch: ai-01 `msg-20260619T231008-8jl3ic` SECONDAIRE #131/#506.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #558, opened 2026-06-20
test(fallacy): pin 8 LinkXxFallback cascade contracts + null-floor discovery

jsboige:
## What

Pin the 8 per-language **`LinkXxFallback`** cascades on `Fallacy` (Fr/En = 2-deep, Ru/Pt/Es/Ar/Fa/Zh = 3-deep: target → En → Fr) — the mind-map link resolution that feeds `FallacyMindMapDocumentConfig.LinkExpression` (`{item.LinkFrFallback}`). **29 contract tests, additive only, no production code change.** Full suite **417/0/5**.

This is the ai-01 **"couverture tests"** runway (dispatched at the 05:08Z tick after merging #555/#556/#557).

## Why

French is the source language. The cascade design is intentionally **asymmetric** — a subtle fragility that had **zero unit coverage** despite being referenced in `MmGeneratorTests` and `MindMapLocalizationRegressionTests`:

- `LinkFrFallback` / `LinkEnFallback` are **2-deep** (Fr↔En: the two source-ish languages).
- `LinkRu/Pt/Es/Ar/Fa/Zh` are **3-deep**: target → En → Fr.

The silent-wrong-output risk: a fallacy that ships with **only a French link** must still resolve to French in every non-source export. A refactor that "symmetrizes" the cascade (e.g. makes every language a 2-deep target→Fr, or reorders En/Fr) would drop the French floor — producing a mind-map node with an empty link, **no exception, no log**, invisible except by opening every generated mind-map.

## Discovery surfaced by writing the tests

The cascade floor is **`null`, not `string.Empty`**. When all three link sources are null, the expression `IsNullOrEmpty(LinkEn) ? LinkFr : LinkEn` returns `LinkFr` which is null. 7 of the initial assertions had to flip from `BeEmpty()` to `BeNull()` — the tests now pin the **real** contract, not an idealized one. This `null`-propagation (empty text in Mustache, but null to any C# consumer) is itself a silent-wrong-output hazard worth documenting.

## The 29 tests (`FallacyLinkFallbackTests`)

| Group | Tests | What it pins |
|---|---|---|
| LinkFrFallback (2-deep) | 3 | primary present → Fr; primary null → En; both null → **null** (documented) |
| LinkEnFallback (2-deep inverse) | 2 | primary present → En; primary null → Fr |
| Non-source × 6 langs (Theory ×4) | 24 | target→En→Fr cascade: target present (short-circuit); target null + En present → En; target+En null + Fr present → Fr (the fragile floor); all null → null |

Reflection helpers (`SetLink`/`GetFallback`) let one `[Theory]` cover all 6 non-source languages uniformly — 4 branches × 6 languages = 24 assertions from 4 methods.

## Verification

- **New tests**: 29/29 pass.
- **Full suite**: **417 passed / 0 failed / 5 skipped** — no regression (baseline 388 + 29 new).
- Build: 0 errors.
- No production code touched — pure additive coverage.

## Related

- Dispatched by: ai-01 (05:08Z tick, "couverture tests" runway, after merging #555/#556/#557, master `d09778ea`).
- Closes the "referenced but unpinned" gap flagged in `MmGeneratorTests` / `MindMapLocalizationRegressionTests`.
- DoD: additive tests only — mergeable without a content gate.

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/Parsing/FallacyLinkFallbackTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/Parsing/FallacyLinkFallbackTests.cs
@@ -1,0 +1,167 @@
+using Argumentum.AssetConverter.Entities;
+using FluentAssertions;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.Parsing
+{
+    /// <summary>
+    /// Contract pin for the 8 <c>LinkXxFallback</c> cascades on <see cref="Fallacy"/> — the
+    /// per-language mind-map link resolution that feeds <c>LinkExpression</c>
+    /// (<c>{item.LinkFrFallback}</c>) in <c>FallacyMindMapDocumentConfig</c>, and which is
+    /// referenced (but NOT behavior-pinned) by <c>MmGeneratorTests</c> and
+    /// <c>MindMapLocalizationRegressionTests</c>.
+    ///
+    /// French is the source language. The cascade design is intentionally ASYMMETRIC — a subtle
+    /// fragility that has had ZERO unit coverage:
+    /// <list type="bullet">
+    /// <item><description><see cref="Fallacy.LinkFrFallback"/> / <see cref="Fallacy.LinkEnFallback"/>
+    /// are 2-deep (Fr↔En: the two source-ish languages).</description></item>
+    /// <item><description><see cref="Fallacy.LinkRuFallback"/> / Pt / Es / Ar / Fa / Zh are 3-deep:
+    /// target → En → Fr (a non-source language falls back to En, then to the French source).</description></item>
+    /// </list>
+    ///
+    /// The silent-wrong-output risk this guards against: if a refactor "symmetrizes" the cascade
+    /// (e.g. makes every language a 2-deep target→Fr, or reorders En/Fr), a fallacy that ships with
+    /// ONLY a French link would resolve to <c>string.Empty</c> in non-EN exports — producing a
+    /// mind-map node with an EMPTY link, no exception, no log. The regression is invisible except
+    /// by opening every generated mind-map and checking each node. These tests pin the cascade
+    /// order additively so such a regression fails loud.
+    /// </summary>
+    public class FallacyLinkFallbackTests
+    {
+        // ─────────────────────────────────────────────────────────────────────────────
+        // (1) LinkFrFallback — 2-deep: Fr → En.
+        // ─────────────────────────────────────────────────────────────────────────────
+
+        [Fact]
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** — #558 (test(fallacy): 8 LinkXxFallback cascade contracts + null-floor discovery)

Verdict: clean additive contract suite, mergeable as-is.

Tests are real assertions, not stubs. All 29 use FluentAssertions on actual `Fallacy.LinkXxFallback` resolution — e.g. `LinkFrFallback_PrimaryNull_FallsBackToEn` asserts `.Should().Be("en-link")` against a `Fallacy` with `LinkFr = null`, and `NonSource_TargetAndEnNull_FrPresent_ReturnsFr` (Theory ×6) asserts the fragile French floor `.Should().Be("fr-link")`. No `Assert.True(true)`, no empty bodies.

8 cascades are coherent and match the body. Fr/En are 2-deep (5 `[Fact]`s: Fr→En primary/fallback/null + En→Fr primary/fallback), the 6 non-source langs (Ru/Pt/Es/Ar/Fa/Zh) are 3-deep via 4 `[Theory]` branches ×6 = 24. Spot-checked the 3-deep chain: when `LinkRu=null`, `LinkEn="en-link"`, the test asserts the fallback returns `"en-link"` (target→En), and when only `LinkFr` is set it asserts `"fr-link"` (the source-language floor) — that's the real target→En→Fr cascade, not a symmetric 2-deep.

Null-floor discovery is genuine and well-documented. The body's claim holds: `IsNullOrEmpty(LinkFr) ? LinkEn : LinkFr` with both null returns `LinkEn` which is null — NOT `string.Empty`. The `LinkFrFallback_BothNull_ReturnsNull` and `NonSource_AllNull_ReturnsNull` (Theory ×6) tests pin this real behavior. The author notes 7 initial assertions had to flip from `BeEmpty()` to `BeNull()` — i.e. the tests were corrected to match the actual contract rather than an idealized one. Worth knowing because that null propagates to Mustache as empty text but stays null to any C# consumer — a silent-wrong-output hazard the contract now documents.

Additive-only confirmed: single new file `FallacyLinkFallbackTests.cs`, +167/-0, zero production `.cs` modified. Class compiles-coherent: proper `using` (xunit, FluentAssertions, Entities), correct `[Fact]`/`[Theory]`/`[InlineData]` attributes, namespace matches folder structure, reflection helpers (`SetLink`/`GetFallback`) are well-formed for the 6-lang Theory fan-out. Security scan clean — no secrets/keys/paths, all test data synthetic (`"fr-link"`, `"xx-link"`).

Comment-only (self-review cap).

PR #561, opened 2026-06-20
test(parse): close ClassMap matrix — TestFallacyCard regression (5 tests)

jsboige:
## What

Add `TestFallacyCardClassMapRegressionTests` — **5 contract tests** that **close the ClassMap matrix at 100%**. **Additive only, no production code change.** Full suite **434/0/5**.

This is the ai-01 **SECONDAIRE** dispatch (`msg-20260620T094434-yj9f75`: "TestFallacyCard ClassMap — dernier gap de la matrice ClassMap, ferme la matrice à 100%").

## Why

`TestFallacyCard` was the **last entity ClassMap without a dedicated regression suite**. The ClassMap matrix is now complete:

| Entity | Coverage |
|---|---|
| Fallacy | `FallacyClassMapRegressionTests` (#485) |
| Rule / ArgumentVirtue / Scenario | `EntityClassMapRegressionTests` (#476) |
| Virtue | `VirtueClassMapRegressionTests` (#559) |
| DnnUiString | `DnnUiStringClassMapRegressionTests` |
| **TestFallacyCard** | **this PR** ✅ |

`TestFallacyCard` is the simplified fallacy entity for visual-test card generation. Its ClassMap declares **5 columns ALL required** (non-`.Optional()`): `Id`, `Title`, `Type`, `IllustrationPath`, `Description`. A header-name drift = silent null properties = empty visual-test cards.

## The 5 tests

| Test | What it pins |
|---|---|
| `MapsAllColumns` | all 5 required columns map to their properties |
| `GetIdReturnsId` | `GetId()` returns `Id` — the image-generator output-dictionary key |
| `RequiredColumnAbsent_Throws` | a required column drift fails LOUD (`HeaderValidationException`), not silent null — the data-loss-prevention contract |
| `ShortRow_DoesNotThrow` | a ragged row yields `string.Empty` for missing fields (`MissingFieldFound` is a log callback, not throw) — mirrors `CsvBaseStrictContractTests` |
| `MultipleRows_LoadInOrder` | multi-row load preserves order |

## Verification

- **New tests**: 5/5 pass.
- **Full suite**: **434 passed / 0 failed / 5 skipped** — no regression (baseline 429 + 5 new).
- Build: 0 errors.
- No production code touched — pure additive coverage.

## Related

- Dispatched by: ai-01 SECONDAIRE (`msg-20260620T094434-yj9f75`).
[...]

PR #562, opened 2026-06-20
test(pdf): #204 extract PdfAuditor page-grid geometry to pure method + 9 tests

jsboige:
## What

Extract the inline A4-based page-grid arithmetic from `PdfAuditor.GetExpectedImageOrder` into a pure, deterministic method `ComputeAuditPageGeometry`, so the audit's layout contract is unit-testable without a PDF render. **Output-neutral** — the call site preserves the exact computation, no behavior change.

Part of the #204 output-neutral extraction lane (cont. po-2024): the matrice ClassMap is closed (#561), SVG disambiguation is pinned (#560). A targeted audit (Explore + code read) found **one real gap left**: the audit computed its page-grid geometry **inline**, with only a code comment guarding the assumption.

## Why

`GetExpectedImageOrder` chunked the deck into page-sized groups using geometry derived from:
- **hardcoded `PageSizes.A4`** (comment: *"Assuming A4 for calculation, needs to be dynamic if possible"*)
- **no page margin** subtracted
- **no header band** reserved

…all of which **diverge from the renderer** (`PrintAndPlayDocument.ComputePageGeometry`, which resolves the actual `docConfig.PageSize` and subtracts margin + header). The audit's `NbCardsPerPage` is therefore systematically **larger** than the renderer's whenever the document is not plain A4-without-header — chunking the deck into pages of the wrong capacity and **silently producing false audit mismatches or false passes**, with no signal beyond the PDF render. This arithmetic had **zero unit coverage**.

## Changes

**`PdfAuditor.cs`** — extraction (output-neutral):
- New `ComputeAuditPageGeometry(cardWidthPoints, cardHeightPoints, configuredNbColumns)` pure static method + `AuditPageGeometry` readonly struct (`NbColumns`, `NbRows`, `NbCardsPerPage`).
- `GetExpectedImageOrder` calls it (same arithmetic, same result).
- XML doc explicitly documents the divergence from the renderer + flags it for a separate behavior-change PR.

**`PdfAuditorPageGeometryContractTests.cs`** — 9 contract tests pinning the real contract:
1. Configured columns honored when `> 0`
2. Zero config → floor(A4w / cardW)
3. Negative config → same fallback (the `> 0` guard)
4. Floor **truncation**, not rounding (the fragile bit — `Math.Round` would regress)
5. Rows = floor(A4h / cardH)
6. Cards-per-page = rows × cols
7. Cards-per-page with configured columns
8. **Hardcoded A4** observable (column count derived from A4, never from actual page size)
9. **Divergence pin** — audit capacity (96) exceeds renderer capacity (84, headered) by the one header row; documents the latent bug

## No-Pendulum

The latent bug is **documented and pinned, NOT fixed**. A behavior change (route the audit through `ComputePageGeometry`, or accept margin/header inputs) belongs in its own PR — this one is strictly output-neutral extraction + contract pinning, matching the lane discipline (#553/#556/#560).

## Verification

- `dotnet test --filter PdfAuditorPageGeometryContractTests` → **9/9 pass**
- Full suite → **438/0/5** (baseline 429 + 9). 0 regression.

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/PdfAuditorPageGeometryContractTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/PdfAuditorPageGeometryContractTests.cs
@@ -1,0 +1,158 @@
+using System;
+using Argumentum.AssetConverter;
+// PdfAuditor is a static class whose namespace shares its name (Argumentum.AssetConverter.PdfAuditor),
+// so a plain PdfAuditor.X call resolves X against the NAMESPACE, not the class. `using static` imports
+// the class's static members directly, letting us call ComputeAuditPageGeometry unqualified.
+using static Argumentum.AssetConverter.PdfAuditor.PdfAuditor;
+using FluentAssertions;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.WebBasedGenerator
+{
+    /// <summary>
+    /// Contract pin for <see cref="PdfAuditor.ComputeAuditPageGeometry"/> — #204 tertiary (cont.
+    /// po-2024): the PDF-audit page-grid geometry contract.
+    ///
+    /// Before chunking the deck into page-sized groups for <see cref="BuildExpectedImageOrder"/>, the
+    /// audit derives how many card columns/rows fit on one sheet and how many cards that sheet holds.
+    /// That capacity drives the page boundaries the expected image sequence is built around — if it is
+    /// wrong, every page boundary is wrong, and the audit silently reports false mismatches or false
+    /// passes against the rendered PDF.
+    ///
+    /// The arithmetic was previously INLINED inside <see cref="PdfAuditor.GetExpectedImageOrder"/> with
+    /// only a code comment ("Assuming A4 for calculation, needs to be dynamic if possible") guarding
+    /// the assumption. It has been extracted (output-neutral — the call site preserves the exact
+    /// computation) into the pure, deterministic <see cref="PdfAuditor.ComputeAuditPageGeometry"/> so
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** Refactor-correctness pass — extraction is behavior-preserving, method is pure, tests are genuine.

- **Output-neutral confirmed.** The moved arithmetic is byte-identical: same `PageSizes.A4` constant, same guards (`configuredNbColumns > 0`), same `(int)` truncation casts, same `nbRows * nbColumns`. Call site passes the exact values (`cardWidthPoints, cardHeightPoints, docConfig.NbColumns`) that were inlined. The only structural change is `nbRows` now surfaces on the struct (previously used only as a local to derive `nbCardsPerPage`) — no runtime difference.
- **Pure method.** `ComputeAuditPageGeometry(float, float, int)` → `AuditPageGeometry` struct: no I/O, no PDF render, no `DateTime.Now`/random, deterministic. Genuinely unit-testable.
- **Tests are real assertions, not tautologies.** Cross-checked `Columns_FloorTruncation`: A4 595.2756/100 = 5.95 → `(int)` truncates to 5 (not 6 with rounding) — a genuine floor-vs-round edge case. `CardsPerPage_IsRowsTimesColumns`: 11×8 = 88, verified manually. The divergence test correctly pins the latent A4-vs-renderer bug as *observable* (audit 96 > renderer 84 with header) without fixing it — appropriate scope for an output-neutral PR.
- Minor: the `using static` workaround for the namespace/class name collision (`Argumentum.AssetConverter.PdfAuditor.PdfAuditor`) is a legitimate C# quirk, well-commented in the test file.

No concerns blocking merge. Good extraction — fragile inline geometry now pinned by 9 contract tests.

PR #564, opened 2026-06-20
test(pdf): #204 pin GetLocalizedFileName contract + document replace-all edge (10 tests)

jsboige:
## What

Add 10 contract tests for `CardSetLocalization.GetLocalizedFileName` — the pure string transform that derives each language's template filename from the default-language one. Part of the #204 output-neutral coverage lane (cont. po-2024).

## Why

`GetLocalizedFileName(fileName, defaultLanguage, targetLanguage)` drives which CardPen template gets loaded for each language: the pipeline localizes the default template path by swapping/inserting the language code. A wrong swap silently loads the wrong template (cards in the wrong language, or no cards). It is a pure, deterministic string transform (no I/O) with **three branches and several subtle edge cases** — and it had **zero unit coverage**.

This was the idle fallback of ai-01's dispatch (`msg-...171324`): *"#204 tests du pur GetLocalizedFileName (low-risk, déjà pure, juste tests)."* Selected after the dispatched Primary (#192 trad quality) and Secondary (#29 memory) had **no actionable surgical target** this tick (see investigation below).

## Contract pinned (10 tests)

1. **Empty/null** → returned as-is (guard clause)
2. **Contains `_{defaultLanguage}`** → replaced with target (happy path: `Argumentum_Fallacies_fr.json` → `..._en.json`)
3. **Path preserved** through replacement (`locales\cards_fr.json` → `locales\cards_pt.json`)
4. **Substring guard** — the match requires a leading `_`, so `french.json` (contains `fr` but not `_fr`) is NOT misread → falls through to insertion (`french_en.json`)
5. **Multiple occurrences — ALL replaced** (documented as-is): `my_fr_cards_fr.json` → `my_en_cards_en.json`. The implementation uses `string.Replace` which swaps every occurrence. A future "replace only the last language tag" would change behavior — flagged for a separate behavior-change PR, not fixed here (No-Pendulum).
6. **No token + extension** → inserted before extension (`cards.json` → `cards_en.json`)
7. **Multi-dot filename** — only the last dot is the extension (`Cards.v2.json` → `Cards.v2_en.json`, keeps the `.v2` infix)
8. **No token + no extension** → appended (`cards` → `cards_en`)
9. **Round-trip neutrality** — swap to target then back restores the original (guards asymmetric swaps)

## Investigation context — why #192 and #29 were skipped this tick

Per the dispatch's deep-queue order, I investigated the Primary (#192 trad quality) and Secondary (#29 memory) **before** falling back to this idle task:

- **#192 (Primaire)** — scanned all translation CSVs (Fallacies/Scenarii/Rules/Virtues × EN/RU/PT) for a measurable quality gap. Result: **no surgical target detectable**. The "12 contaminated example_pt cells" surfaced by the initial scan were false positives (proper nouns `Joana d'Arc`/`La Fontaine`, FR↔PT lexical coincidences like `mais`/`est`). Strict scan for impossible-in-PT French elisions (`l'`/`qu'`/`c'est`) → **0 cells**. Exact FR==target copies → **0** everywhere. The #216 contamination bug was correctly fixed; residual quality is polish requiring **native-speaker judgment** (not a grep-able gap) — gated to jsboige/locuteurs per the dispatch.
- **#29 (Secondaire)** — confirmed **#436 (deterministic Magick.NET dispose, P0 memory fix) is MERGED and in place** (`PdfManager.GeneratePdfsFromImages` uses `using var collection`). The P0 fix is delivered; #29 is closable with this proof. A dispose regression test is feasible but fragile to design (MagickImageCollection dispose isn't publicly observable; WeakReference/GC tests are flaky) — deferred rather than rushed.

Falling back to the clean idle task (this PR) rather than forcing a fragile/incorrect fix respects No-Pendulum. Both findings are reported to ai-01 for lane redirection.

## Verification

- `dotnet test --filter LocalizedFileNameContractTests` → **10/10 pass**
- Full suite → **453/0/5** (baseline 443 + 10). 0 regression.

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/LocalizedFileNameContractTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/LocalizedFileNameContractTests.cs
@@ -1,0 +1,144 @@
+using Argumentum.AssetConverter;
+using FluentAssertions;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.WebBasedGenerator
+{
+    /// <summary>
+    /// Contract pin for <see cref="CardSetLocalization.GetLocalizedFileName"/> — #204 idle (cont.
+    /// po-2024): the localized-filename transform contract.
+    ///
+    /// When the pipeline localizes a CardSet, it derives the output template filename from the
+    /// default-language one by swapping (or inserting) the language code. That derived filename
+    /// drives which template gets loaded for each language — if it is wrong, the wrong template (or
+    /// none) is loaded, silently producing cards in the wrong language or no cards at all.
+    ///
+    /// <see cref="CardSetLocalization.GetLocalizedFileName"/> is a pure, deterministic string
+    /// transform (no I/O) with three branches and several subtle edge cases (substring matches,
+    /// multiple occurrences, multi-dot filenames) that had ZERO unit coverage. These tests pin the
+    /// contract additively so a refactor cannot silently change which template a language resolves to.
+    /// </summary>
+    public class LocalizedFileNameContractTests
+    {
+        // ─────────────────────────────────────────────────────────────────────────────
+        // (1) EMPTY / NULL — passed through untouched (guard clause).
+        // ─────────────────────────────────────────────────────────────────────────────
+
+        [Theory]
+        [InlineData(null)]
+        [InlineData("")]
+        public void EmptyOrNull_ReturnedAsIs(string fileName)
+        {
+            // The first guard returns the input verbatim, so no language substitution is attempted.
+            CardSetLocalization.GetLocalizedFileName(fileName, "fr", "en")
+                .Should().Be(fileName, "an empty/null filename cannot be localized and is returned as-is.");
+        }
+
+        // ─────────────────────────────────────────────────────────────────────────────
+        // (2) CONTAINS _{defaultLanguage} — replaced (the happy path for standard templates).
+        // ─────────────────────────────────────────────────────────────────────────────
+
+        [Fact]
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** — COMMENT (LGTM)

Test-only (+144/-0): adds 10 contract tests for `CardSetLocalization.GetLocalizedFileName` (pure string transform, #204 output-neutral lane). Verified each expected value against the **actual implementation** on `master` (`CardSetLocalization.cs`) — all 8 visible tests match real behavior exactly:

- Guard: null/"" → returned as-is (returns before any `.Contains`, so null is safe). ✓
- Branch 1 `Contains($"_{defaultLanguage}")` → `Replace`: `Argumentum_Fallacies_fr.json`→`_en.json`, `locales\cards_fr.json`→`_pt.json` (path preserved). ✓
- Substring guard: `french.json` (contains "fr" not "_fr") → falls through to insert → `french_en.json`. ✓ pins the `_`-prefix requirement.
- Multi-occurrence: `my_fr_cards_fr.json` → `my_en_cards_en.json` (string.Replace swaps ALL) — honestly documented as the fragile bit, behavior pinned as-is. ✓
- Branch 2 insert-before-ext: `cards.json`→`cards_en.json`; multi-dot `Cards.v2.json`→`Cards.v2_en.json` (Path.GetExtension treats last dot only). ✓
- Branch 3 append (no ext): `cards`→`cards_en`. ✓

Genuine contract tests (FluentAssertions + xUnit), not tautologies — they pin real branches and subtle edges (substring guard, multi-occurrence `Replace`, multi-dot `Path` semantics). Additive (no production code changed), deterministic string inputs, no secrets. LGTM clean. — NanoClaw (myia-ai-01)

PR #570, opened 2026-06-21
test(ontology): #204 pin OwlDocumentConfig.GetId IRI-fragment transform (10 tests)

jsboige:
## What

Pin the OWL IRI-fragment transform contract for `OwlDocumentConfig.GetId(string)` — a pure, deterministic helper surfaced by the #204 coverage sweep with **zero isolated coverage**.

## Why

The OWL generator turns each fallacy's display name into the fragment identifier of its concept IRI (e.g. `https://www.argumentum.games/...#AdHominem`). That transform is `GetId`: Humanizer `Camelize()` (camelCase join) then strip apostrophes, hyphens and commas — characters forbidden in IRI fragments. A bug here produces duplicate or invalid concept IRIs, which silently corrupts the generated ontology (collision, or a fragment that breaks IRI resolution). It is pure & deterministic — no I/O, no config state — yet the E2E ontology tests exercise it only indirectly. These tests pin the contract additively.

## Tests added (10)

All exact-match against real Humanizer 2.14.1 output (captured via a standalone probe — see note below):

- **Camelize joins words camelCase** (4 Theory cases): `"Ad Hominem"→"adHominem"`, `"straw man"→"strawMan"`, `"red herring"→"redHerring"`, `"Hasty Generalization"→"hastyGeneralization"`
- **Apostrophes stripped**: `"L'appel"→"lappel"` (French elisions don't leak into the IRI)
- **Hyphens stripped**: `"A-B"→"aB"` (compound names produce a valid fragment)
- **Commas stripped**: `"A, B"→"aB"` (list-style names produce a valid fragment)
- **End-to-end French fallacy name**: `"Appel à l'autorité"→"appelÀLautorité"`
- **Determinism** — pure function, identical inputs → identical fragments
- **Generic invariant** — across 8 inputs (incl. accented + compound), the fragment never contains `'`, `-`, or `,`

## Finding pinned (not fixed)

Humanizer `Camelize()` **preserves accented characters** (it does not ASCII-fold): `"à"` survives and is uppercased to `"À"` at the segment start. This is the **observed contract** — the fragment is apostrophe/hyphen/comma/space-free but may contain accented letters. Pinned exactly so a future Humanizer upgrade (or an added ASCII-fold) that changes accent handling is caught. Accented IRI fragments are technically valid (IRIs allow non-ASCII per RFC 3987); if strict-ASCII URIs were ever required, `GetId` is the place that would change. Acted on separately if at all — this PR is output-neutral, additive tests only (no production code changed).

## Verification

- `dotnet test --filter OwlGetIdPureContractTests` → **10/10 pass**
- Full suite → **479/0/5** (baseline 469 + 10). 0 regression.

## Probe note

Humanizer `Camelize()` output is non-obvious (camelCase, not PascalCase; accents preserved, not folded). I captured the real outputs from Humanizer 2.14.1 via a standalone console probe before writing the assertions, so the expected values are observed behavior — not assumptions.

Continues the #204 lane per ai-01's dispatch ("enchaîne deep-queue #204 ; pas de stall").

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlGetIdPureContractTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/Ontology/OwlGetIdPureContractTests.cs
@@ -1,0 +1,130 @@
+using Argumentum.AssetConverter.Ontology;
+using FluentAssertions;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.Ontology
+{
+    /// <summary>
+    /// Contract pin for <see cref="OwlDocumentConfig.GetId"/> — #204 coverage sweep (cont. po-2024):
+    /// the OWL IRI-fragment transform contract.
+    ///
+    /// The OWL generator turns each fallacy's display name into the fragment identifier of its concept
+    /// IRI (e.g. <c>https://www.argumentum.games/...#AdHominem</c>). That transform is
+    /// <see cref="OwlDocumentConfig.GetId"/>: Humanizer <c>Camelize()</c> (PascalCase join) then strip
+    /// apostrophes, hyphens and commas — characters forbidden in IRI fragments. A bug here produces
+    /// duplicate or invalid concept IRIs, which silently corrupts the generated ontology (collision,
+    /// or a fragment that breaks IRI resolution). It is pure &amp; deterministic — no I/O, no config
+    /// state — yet had ZERO isolated coverage (the E2E ontology tests exercise it only indirectly).
+    /// These tests pin the contract additively.
+    /// </summary>
+    public class OwlGetIdPureContractTests
+    {
+        // ─────────────────────────────────────────────────────────────────────────────
+        // (1) Camelize — Humanizer Camelize() produces camelCase (first segment lowercased, subsequent
+        //     segments kept/uppercased-on-first-letter) and joins the words into a single token. The
+        //     exact outputs below were captured from the real Humanizer 2.14.1 transform.
+        // ─────────────────────────────────────────────────────────────────────────────
+
+        [Theory]
+        [InlineData("Ad Hominem", "adHominem")]
+        [InlineData("straw man", "strawMan")]
+        [InlineData("red herring", "redHerring")]
+        [InlineData("Hasty Generalization", "hastyGeneralization")]
+        public void Camelize_JoinsWordsCamelCase_NoSpaces(string input, string expected)
+        {
+            OwlDocumentConfig.GetId(input).Should().Be(expected,
+                "Camelize() joins the words into a single camelCase token, dropping the spaces.");
+        }
+
+        // ─────────────────────────────────────────────────────────────────────────────
+        // (2) Apostrophes stripped — French elisions (l', d', qu') would otherwise land in the IRI.
+        // ─────────────────────────────────────────────────────────────────────────────
+
+        [Fact]
+        public void Apostrophes_Stripped()
+        {
+            // "L'appel" → Camelize "l'appel" → apostrophe stripped → "lappel".
+            OwlDocumentConfig.GetId("L'appel").Should().Be("lappel",
+                "apostrophes are stripped after Camelize so French elisions don't leak into the IRI fragment.");
+        }
+
+        // ─────────────────────────────────────────────────────────────────────────────
+        // (3) Hyphens stripped — compound names ("A-B") would create an invalid fragment otherwise.
+        // ─────────────────────────────────────────────────────────────────────────────
+
+        [Fact]
+        public void Hyphens_Stripped()
+        {
+            // "A-B" → Camelize "a-b" → hyphen stripped → "aB".
+            OwlDocumentConfig.GetId("A-B").Should().Be("aB",
+                "hyphens are stripped so compound names produce a valid IRI fragment.");
+        }
+
+        // ─────────────────────────────────────────────────────────────────────────────
+        // (4) Commas stripped.
+        // ─────────────────────────────────────────────────────────────────────────────
+
+        [Fact]
+        public void Commas_Stripped()
+        {
+            // "A, B" → Camelize "a, b" → comma stripped → "aB".
+            OwlDocumentConfig.GetId("A, B").Should().Be("aB",
+                "commas are stripped so list-style names produce a valid IRI fragment.");
+        }
+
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** — test review · Argumentum #570 (OwlGetId contract tests)
**Verdict:** LGTM (0 blocking concerns; 2 framing nits)
**Scope:** 1 file, +130/-0, test-only, 10 contract tests pinning `OwlDocumentConfig.GetId` — the name→IRI-fragment transform feeding the OWL concept IRI generator.
**Tests vs SUT:** All expected values verified correct against the real SUT. The method (in `OwlGeneratorConfig.cs`, class `OwlDocumentConfig`) is:
```csharp
public static string GetId(string text) =>
    text.Camelize().Replace("'","").Replace("-","").Replace(",","");
```
I pulled the actual Humanizer **2.14.1** source (`Pascalize` = `Regex.Replace(input, "(?:^|_| +)(.)", m => m.Groups[1].Value.ToUpper())`; `Camelize` = Pascalize with first char lowercased) and traced every test case by hand:
- `Camelize_JoinsWordsCamelCase_NoSpaces` (4 InlineData): `"Ad Hominem"→adHominem`, `"straw man"→strawMan`, `"red herring"→redHerring`, `"Hasty Generalization"→hastyGeneralization` — all match. Pascalize uppercases the char after `^` and after ` +`, Camelize lowercases the first.
- `Apostrophes_Stripped`: `"L'appel"` → Pascalize `"L'appel"` (apostrophe is not a word-boundary in the regex) → Camelize `"l'appel"` → strip `'` → `"lappel"` ✓. The comment trace in the test (`Camelize "l'appel"`) is accurate.
- `Hyphens_Stripped`: `"A-B"` → Pascalize `"A-B"` (`-` is not a regex boundary) → Camelize `"a-B"` → strip `-` → `"aB"` ✓.
- `Commas_Stripped`: `"A, B"` → Pascalize `"A,B"` → Camelize `"a,B"` → strip `,` → `"aB"` ✓.
- `FrenchFallacyName_ProducesFragment_AccentsPreserved` (the load-bearing one): `"Appel à l'autorité"` → Pascalize `"AppelÀL'autorité"` (the `à` after the space gets uppercased to `À` via `.ToUpper()`, which is culture/Unicode-aware and correctly produces `À`) → Camelize `"appelÀL'autorité"` → strip `'` → `"appelÀLautorité"` ✓. This is the strongest correctness signal — the expected value is non-obvious and genuinely pinned.
- `Deterministic_SameInput_SameOutput`: trivially true for a pure string transform; legitimate characterization guard.
- `Result_NeverContainsForbiddenChars`: traced all 8 loop inputs (`"Ad Hominem"`, `"L'appel à l'autorité"`, `"A-B, C"`, `"red-herring"`, `"tu, quoque"`, `"Slippery Slope"`, `"post hoc"`, `"Ménage à trois"`) — every output is free of `'`/`-`/`,` ✓. Note `"Ménage à trois"→ménageÀTrois` exercises accent preservation in the loop, not just the dedicated Fact.

**Genuineness:** Real. Every test calls the real static `OwlDocumentConfig.GetId(string)` with concrete string inputs and asserts concrete outputs via FluentAssertions `.Should().Be(...)`. No `[Moq]`/`Mock<OwlDocumentConfig>`, no `Assert.True(true)`, no test re-asserts its own input. The forbidden-chars test is a property check over a corpus, not a tautology.

**Coverage:** Spans the transform's branches well for the documented contract (Camelize join, apostrophe, hyphen, comma, accents, determinism, forbidden-chars invariant). The method's only 3 distinct strip operations (`'`, `-`, `,`) each get a dedicated Fact plus the loop. Honest gaps (not invented requirements): (a) no test for empty input `""` — `GetId("")` returns `""` (Pascalize of empty is empty, length-guard short-circuits) — harmless but a one-liner characterization test would lock the null/empty edge; (b) no test for a name that is *already* a bare id (`"fooBar"` → `"fooBar"`) — Pascalize's regex only fires on `^` so first char gets uppercased then re-lowercased by Camelize, returning the input unchanged, which is worth pinning. Neither is blocking.

**Verified:**
- Test class compiles: `using Argumentum.AssetConverter.Ontology;` + `using FluentAssertions;` + `using Xunit;` matches sibling files (`OwlAdapterRegressionTests.cs`, `OwlAdapterSurvivorFallbackTests.cs`) using the same triple; test project csproj pulls `FluentAssertions 8.5.0`, `xunit 2.9.2` ✓.
- `OwlDocumentConfig.GetId` is called as a static method, matching the SUT signature `public static string GetId(string text)` ✓.
- Humanizer 2.14.1 pinned in the production csproj (`Argumentum.AssetConverter.csproj` → `<PackageReference Include="Humanizer" Version="2.14.1" />`), matching the XML doc's "captured from the real Humanizer 2.14.1 transform" claim. This matters: `Camelize` behavior changed materially in 3.x (the 3.x source has `TryPascalizeAscii` which ASCII-folds and would mangle the accent case), so pinning to 2.14.1 is load-bearing for test #5's correctness. A Humanizer bump to 3.x would break the accent test — which is exactly the kind of regression a characterization pin should catch. Good.
- Security scan: no secrets/tokens/keys in the file (the two `grep` hits on "token" are in a FluentAssertions reason string and a doc comment — not credentials).
- Anti-théâtre (#1019): legitimate. The PR title says "pin ... transform" and these are genuine characterization tests over behavior that demonstrably exists in the SUT today (not aspirational). The accent test in particular documents a real, non-obvious contract worth locking before any refactor.

**Concerns (non-blocking):**
1. **PR-description framing is misleading, but the test file's own XML doc is correct.** The task/PR description calls `GetId` an "IRI-fragment extraction transform (OWL IRIs → local fragment/id)" and points at a file path `Ontology/OwlDocumentConfig.cs`. Reality: `GetId` *builds* a fragment from a **display name** (it does not parse an IRI), and the class `OwlDocumentConfig` lives inside `Ontology/OwlGeneratorConfig.cs` (no standalone file). The XML doc comment in the test file itself is accurate ("turns each fallacy's display name into the fragment identifier"). No code impact, but a reader following the PR description would look for the wrong behavior in the wrong file.
2. **`Result_NeverContainsForbiddenChars` corpus omits a few edge shapes** — e.g. a name with a *trailing* apostrophe (`"foo'"`) or a double hyphen (`"A--B"`). These still clean correctly under the `Replace` chain but aren't exercised. Optional addition, not a defect.

— NanoClaw

PR #571, opened 2026-06-21
test(localization): #204 pin LocalizationConfig.BuildLanguageList language-set contract (7 tests)

jsboige:
## What

Pin the language-set decision contract for `LocalizationConfig.BuildLanguageList` — a pure, deterministic method with **zero unit coverage** that decides which languages the entire asset-generation pipeline processes.

## Why

`BuildLanguageList` is called from `HarvestManager` (per-CardSet, [HarvestManager.cs:88](Generation/Converters/Argumentum.AssetConverter/WebBasedGenerator/Cardpen/HarvestManager.cs#L88)) and `ParallelDocumentCreatorConfigBase` (per-mind-map, [ParallelDocumentCreatorConfigBase.cs:53](Generation/Converters/Argumentum.AssetConverter/ParallelDocumentCreatorConfigBase.cs#L53)), each driving a `Parallel.ForEachAsync` over the returned list. So the return value is the **single decision point for "how many localized outputs ship for this CardSet / mind-map"**. A regression here silently changes the output count per locale — e.g. a whole locale stops producing cards/PDFs with no error message.

The contract has three observable branches, each a silent-corruption hazard:
- **`Enabled == false`** → only `DefaultLanguage` (the localization gate — prevents empty/partial outputs for locales whose templates/data aren't ready)
- **`Enabled == true`** → `DefaultLanguage` **first**, then every translation's `targetLanguage` (dropping the default loses the base output; dropping the `AddRange` loses every translation)
- **Default language always first** (call sites rely on this for harvest caching/overwrite ordering)

## Purity note

The method reads only two in-memory mutable config properties (`DefaultLanguage`, `Enabled`) and the input list — no File/HTTP/Playwright I/O, no randomness, no time. An earlier sweep (#567) flagged it "config-dependent" and skipped it; that is **overly cautious** — config *state* is not I/O, and the method is pure w.r.t. its inputs, trivially testable by constructing a `LocalizationConfig` with controlled property values. Confirmed zero coverage (no reference to `BuildLanguageList` anywhere in the test project).

## Tests added (7)

All output-neutral, additive (no production code changed):

1. **Disabled** → returns only default, ignores translations (gate)
2. **Enabled** → default-first then targets in order (happy path)
3. **Default always first, no de-duplication** — even if a target repeats the default. Subtle: the implementation prepends the default verbatim and appends targets verbatim — an added `Distinct()` would change the output count and is caught
4. **Enabled + empty translations** → still yields default (never empty list → never zero outputs)
5. **Custom default honored** (not hardcoded to `"fr"`)
6. **Disabled + custom default** → returns only that custom default (gate + seed independent)
7. **Determinism** — same inputs → same list

## Verification

- `dotnet test --filter BuildLanguageListContractTests` → **7/7 pass**
- Full suite → **476/0/5** (baseline 469 + 7). 0 regression.

## SDDD note (verify-before-code)

The Explore audit also surfaced `CsvBase.RemoveDiacritics` as a top candidate — I **declined it**. Verify-before-code found it is already covered end-to-end by `EntityClassMapRegressionTests` (Scenario accented-header regression, lines 175-193: `coordonnées`/`catégorie`/`sous-catégorie` fed through the real `LoadFromContent` and confirmed to resolve). Pinning it again would have been make-work (No-Pendulum). `BuildLanguageList` was the genuine gap.

Continues the #204 lane per ai-01's dispatch ("enchaîne deep-queue #204 ; pas de stall").

### DIFF
--- a/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/BuildLanguageListContractTests.cs
+++ b/Generation/Converters/Argumentum.AssetConverter.Tests/WebBasedGenerator/BuildLanguageListContractTests.cs
@@ -1,0 +1,202 @@
+using System.Collections.Generic;
+using Argumentum.AssetConverter;
+using FluentAssertions;
+using Xunit;
+
+namespace Argumentum.AssetConverter.Tests.WebBasedGenerator
+{
+    /// <summary>
+    /// Contract pin for <see cref="LocalizationConfig.BuildLanguageList"/> — #204 coverage sweep
+    /// (cont. po-2024): the language-set decision contract for the entire asset-generation pipeline.
+    ///
+    /// <see cref="LocalizationConfig.BuildLanguageList"/> decides WHICH languages the pipeline
+    /// processes. It is called from <c>HarvestManager</c> (per-CardSet) and
+    /// <c>ParallelDocumentCreatorConfigBase</c> (per-mind-map), each driving a
+    /// <c>Parallel.ForEachAsync</c> over the returned list — so the return value is the single
+    /// decision point for "how many localized outputs ship for this CardSet / mind-map".
+    ///
+    /// The contract has three observable branches, each a silent-corruption hazard if regressed:
+    /// <list type="bullet">
+    /// <item><c>Enabled == false</c> → returns ONLY <see cref="LocalizationConfig.DefaultLanguage"/>
+    /// (the base language, default "fr"). A regression that ignored the gate would leak target
+    /// languages even when localization is off, producing empty/partial outputs for locales whose
+    /// templates or data aren't ready.</item>
+    /// <item><c>Enabled == true</c> → returns <c>DefaultLanguage</c> FOLLOWED BY every
+    /// translation's <c>targetLanguage</c>. A regression that dropped the default would lose the
+    /// base-language output; one that dropped the <c>AddRange</c> would lose every translation.</item>
+    /// <item>The default language is ALWAYS first, even when enabled — base language is generated
+    /// before translations (call sites rely on this ordering for harvest caching/overwrite).</item>
+    /// </list>
+    ///
+    /// The method reads only two in-memory mutable config properties (<c>DefaultLanguage</c>,
+    /// <c>Enabled</c>) and the input list — no File/HTTP/Playwright I/O, no randomness, no time —
+    /// so it is deterministic given the object state and trivially unit-testable by constructing a
+    /// <see cref="LocalizationConfig"/> with controlled property values. (An earlier sweep flagged
+    /// it "config-dependent"; that is overly cautious — config *state* is not I/O, and the method
+    /// is pure with respect to its inputs.) It had ZERO unit coverage. These tests pin the contract
+    /// additively (no production code changed).
+    /// </summary>
+    public class BuildLanguageListContractTests
+    {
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** — test review · Argumentum #571 (LocalizationConfig.BuildLanguageList contract tests)
**Verdict:** LGTM (0 blocking concerns; 1 optional edge-nit)
**Scope:** 1 file, +202/-0, test-only. New `BuildLanguageListContractTests.cs` — 7 contract tests pinning `LocalizationConfig.BuildLanguageList`, the single decision point for which languages the asset-generation pipeline processes. No production code changed (additive). base `master`, mergeable=clean.

**SUT + trace:** All 7 expected values verified against the real method (`Localization/LocalizationConfig.cs:47-55`):
```csharp
public List<string> BuildLanguageList(List<(string sourceLanguage, string targetLanguage)> translations)
{
    var targetLanguages = new List<string>(new[] { DefaultLanguage });
    if (Enabled)
        targetLanguages.AddRange(translations.Select(t => t.targetLanguage));
    return targetLanguages;
}
```
- (1) Disabled gate → `["fr"]` only (AddRange skipped) ✓
- (2) Enabled happy path → `["fr","en","ru","pt"]` (default + targets in order) ✓
- (3) **Load-bearing:** default + a translation targeting the same code → **`["fr","fr","en"]`, no de-dup**. Traced correct — the SUT prepends `DefaultLanguage` verbatim then `AddRange`s every `targetLanguage` verbatim with no `Distinct()`, so "fr" appears twice. This is the non-obvious contract worth pinning (an added `Distinct()` would silently change the output count and thus the number of generated localized outputs). ✓
- (4) Enabled + empty translations → `["fr"]` (seed is unconditional) ✓
- (5) Custom default `"en"` honored (not hardcoded `"fr"`) → `["en","es","ar"]` ✓
- (6) Disabled + custom default `"zh"` → `["zh"]` (gate + seed independent) ✓
- (7) Determinism (pure method, no mutation/randomness) ✓

**Call-site claims verified:** the XML doc asserts the method is called from `HarvestManager` (per-CardSet) and `ParallelDocumentCreatorConfigBase` (per-mind-map) — code search confirms `BuildLanguageList` appears in both. The "reads only `DefaultLanguage` + `Enabled`, pure w.r.t. inputs" claim is confirmed by the SUT (no File/HTTP/time).

**Genuineness:** Real. Every test constructs a `LocalizationConfig` with concrete property values and asserts concrete outputs via FluentAssertions `.Should().Equal(...)`. No mocks, no `Assert.True(true)`, no tautology. Characterization pins over behavior that demonstrably exists today.

**Security:** clean (test-only, no secrets/keys; FluentAssertions reason strings are the only string content).

**Anti-théâtre (#1019):** legitimate — these pin a real, load-bearing contract (the no-de-dup behavior in particular is a silent-corruption hazard if regressed, exactly as the doc argues).

**Nit (non-blocking, optional edge):** no test for `translations == null` (Enabled + null → `NullReferenceException` on `.Select`) nor for `DefaultLanguage` null/empty. If the contract intends to tolerate null inputs that's worth pinning; if not, a one-liner characterization test documenting the NRE is the honest pin. Neither is blocking — the current 7 cover the documented contract fully.

— NanoClaw

